PR #6, opened 2022-03-23
fix(core/supervisor): don't restart actor if group is terminating

GoldsteinE:


loyd:
It seems that the logic has been implemented already. Close?

GoldsteinE:
Yes, closing this.

PR #56, opened 2022-04-29
feat(dumper): use human-friendly type for max_dump_size

l4l:
So in config one may write `5 MiB` instead of `5242880`


loyd:
Outdated, now `bytesize` provides `serde` support

PR #77, opened 2022-06-24
fix(elfo-test): drive timers to completion

Leonqn:
Previous version actually doesn't work (you can see it in stuck tests). We need to drive timers to completion in separate thread.

### DIFF
--- a/elfo-test/src/proxy.rs
+++ b/elfo-test/src/proxy.rs
@@ -12,7 +12,7 @@ use std::{
 
 use futures_intrusive::{
     channel::shared,
-    timer::{LocalTimer, LocalTimerService, StdClock},
+    timer::{LocalTimer, StdClock, TimerService},
 };
 use once_cell::sync::Lazy;
 use serde::{de::Deserializer, Deserialize};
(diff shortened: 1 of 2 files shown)


elfo-test/src/proxy.rs
```diff
@@ -88,15 +88,25 @@ impl Proxy {
     #[track_caller]
     pub fn recv(&mut self) -> impl Future<Output = Envelope> + '_ {
         static STD_CLOCK: Lazy<StdClock> = Lazy::new(StdClock::new);
+        static TIMER_SERVICE: Lazy<Arc<TimerService>> = Lazy::new(|| {
+            let timer_service = Arc::new(TimerService::new(&*STD_CLOCK));
+            thread::spawn({
```
loyd, line 93: Maybe it's better just to spawn a task?

Leonqn: And how make it non busy loop task as we can't use tokio::sleep here?

loyd: You're right.

PR #81, opened 2022-09-07
feat(macros): expose macro internals

GoldsteinE:
Expose internals of `elfo-macros` in a separate `elfo-macros-impl` crate, so downstream crates can define their own version of macros with modified path to `elfo`.


elfo-macros-impl/src/lib.rs
```diff
@@ -0,0 +1,4 @@
+extern crate proc_macro;
+
+pub mod message;
+pub mod msg;
```
loyd, line 4: It's better to expose only `message_impl` and `msg_impl` without exposing the internal structure of modules, which can be changed in the future.

elfo-macros-impl/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "elfo-macros-impl"
+version = "0.1.14"
```
loyd, line 3: 0.1.0 for new crates, there is no coordination between them for now

loyd (changes requested):

loyd (approved):

PR #86, opened 2023-04-12
feat: provide Clone for ActorMeta

GoldsteinE:


loyd:
It's preferable to use `Arc<ActorMeta>` everywhere, so I will close the issue for now.

PR #90, opened 2023-05-29
fix(configurer): send UpdateConfig as a regular message instead of a request

laplab:


elfo-configurer/src/lib.rs
```diff
@@ -210,30 +210,29 @@ impl Configurer {
...
-                wrap_request_future(fut, group)
+                wrap_long_running_future(
+                    fut,
+                    group,
+                    "group is validing the config suspiciously long",
+                )
             })
             .collect::<Vec<_>>();
 
-        // TODO: use `try_join_all`.
```
laplab, line 236: I am not sure `try_join_all` is the best choice here. According to the [documentation](https://docs.rs/futures/latest/futures/future/fn.try_join_all.html):

```
If any future returns an error then all other futures will be **canceled**...
```

Cancelling a `ValidateConfig` request may result in a situation where only subset of errors are returned to the user. This in turn may lead to the user performing multiple round trips of updating the configs, each time resolving new part of the bigger problem.

What do you think?

elfo-configurer/src/lib.rs
```diff
@@ -244,7 +243,6 @@ impl Configurer {
                 Ok(Ok(_)) | Err(_) => None,
                 Ok(Err(reject)) => Some((group, reject.reason)),
             })
-            // TODO: provide more info.
```
laplab, line 247: Just wanted to ask, what kind of info do you think we should provide here? I am fine to keep the comment, just want it to be more specific.

elfo-configurer/src/lib.rs
```diff
@@ -255,15 +253,65 @@ impl Configurer {
...
+
+                wrap_long_running_future(
+                    fut,
+                    group,
+                    "some actors in the group have congested mailboxes, config update is stalled",
+                )
+            })
+            .collect::<Vec<_>>();
+
+        let errors = future::join_all(futures)
```
laplab, line 282: I think that `try_join_all` is also undesirable here because we might end up in a situation where half of actors applied the config successfully, then just one errored and caused the other half to have an old config.

CHANGELOG.md
```diff
@@ -9,12 +9,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 - **BREAKING** supervisor: actors discard `ValidateConfig` by default ([#87]).
+- **BREAKING** configurer: send `UpdateConfig` as a regular message instead of a request ([#90]).
```
laplab, line 12: Need to double-check if this is breaking or not. I am not sure if the request token is still being generated if the message of type `Request` is send as a regular message. If not, this might break existing actors.

laplab: @loyd I read through the code a little bit more, we might need to either change `UpdateConfig` from request to a regular message or update `msg!` to allow matching without token for request messages + special handling on the side of the requested actor to discard unmonitored replies to requests sent as a regular message.

Changing `UpdateConfig` type sounds like a way to go for me, curious what you think about it.

elfo-configurer/src/lib.rs
```diff
@@ -255,15 +253,65 @@ impl Configurer {
...
+            .await
+            .into_iter()
+            .filter_map(|(group, result)| match result {
+                Ok(_) => None,
+                Err(err) => {
```
laplab, line 287: Should we ignore errors altogether here, like the current code does? It is valid for the actor to not accept any config updates at all.

elfo/Cargo.toml
```diff
@@ -46,6 +46,7 @@ tracing = "0.1.25"
 metrics = "0.17"
 tracing-subscriber = "0.3"
 serde = { version = "1.0.120", features = ["derive"] }
+serde-value = "0.7.0"
```
loyd, line 49: ?

elfo/tests/update_config.rs
```diff
@@ -0,0 +1,83 @@
+#![cfg(feature = "test-util")]
+
+use elfo::{config::AnyConfig, messages::ConfigUpdated, prelude::*};
+use elfo_core::messages::{ConfigRejected, UpdateConfig};
```
loyd, line 4: Integration tests in `elfo/tests` shouldn't use `elfo_core`, only `elfo`, because it's a facade.

elfo/tests/update_config.rs
```diff
@@ -0,0 +1,83 @@
...
+    let config = toml! {
+        limit = 128
+    };
+    let proxy = elfo::test::proxy(blueprint, config).await;
+
+    proxy.send(StartSingleton).await;
+    assert_eq!(proxy.request(GetLimit).await, 128);
+
+    // Update with valid config through message.
+    let config = AnyConfig::new(
```
loyd, line 45: I believe it should be `AnyConfig::deserialize(toml! {..})`.

Moreover, It would be nice to rename `fn new` to `from_value` and mark it as `#[stability::unstable]` because it's an implementation detail

elfo-core/src/supervisor.rs
```diff
@@ -191,6 +191,7 @@ where
                 Err(reason) => {
                     msg!(match envelope {
                         (messages::UpdateConfig { .. }, token) => {
+                            error!(group = %self.meta.group, %reason, "invalid config");
```
loyd, line 194: `self.in_scope(..)`. Supervisors run in the scope of a caller.

loyd: Also add `, ignoring` to tell users what's going on.

elfo-configurer/src/lib.rs
loyd: You're right and it's exactly why `try_join_all` isn't used.

elfo-configurer/src/lib.rs
loyd: Actor keys

elfo/tests/update_config.rs
laplab: Yep, it's rust-analyzer's shenanigans

PR #92, opened 2023-06-13
refactor(logger): configure targets in logger config

GoldsteinE:


elfo-core/src/logging/filter.rs
```diff
@@ -1,55 +1,32 @@
-use tracing::{level_filters::LevelFilter, Metadata, Subscriber};
-use tracing_subscriber::{filter::Targets, layer::Context, Layer};
+use tracing::{level_filters::LevelFilter, Metadata};
```
loyd, line 1: The whole `LogFilter` can be removed now because enabled levels are already checked by
```rust
            if !scope.permissions().is_logging_enabled(level) {
                return false;
            }
```
in the filtering layer.

elfo-logger/src/lib.rs
```diff
@@ -54,32 +54,31 @@ struct PreparedEvent {
 }
 
 // TODO: revise factory (return also `FilteringLayer` somehow).
-pub fn new() -> (PrintingLayer, Blueprint) {
+pub fn new() -> (PrintingLayer, FilteringLayer, Blueprint) {
```
loyd, line 57: Let's make it private for now.

elfo-logger/src/config.rs
```diff
@@ -27,3 +38,32 @@ pub(crate) struct Format {
     pub(crate) with_module: bool,
     // TODO: colors
 }
+
+// REVIEW: deduplicate with core
```
loyd, line 42: `TODO` or `XXX`

elfo-logger/src/filtering_layer.rs
```diff
@@ -1,46 +1,98 @@
...
-pub(crate) struct FilteringLayer {
-    other_max_level: Level,
+struct FilteringConfig {
+    targets: Targets,
+}
+
+impl Default for FilteringConfig {
+    fn default() -> Self {
+        Self {
+            targets: Targets::new().with_default(LevelFilter::INFO),
```
loyd, line 21: I believe it should be `TRACE`, otherwise it overrides actor levels until the logger is configured.

elfo-logger/src/filtering_layer.rs
```diff
@@ -1,46 +1,98 @@
...
                 CheckResult::NotInterested => false,
                 CheckResult::Limited => {
                     stats::counter_per_level("elfo_limited_events_total", level);
                     false
                 }
             }
         })
-        // TODO: limit events outside the actor system?
-        .unwrap_or_else(|| *meta.level() <= self.other_max_level)
+        .unwrap_or(targets_enabled)
```
loyd, line 94: It shouldn't rely on the configuration of `targets_enabled` and just use `INFO` here.

loyd (changes requested):

elfo-logger/src/filtering_layer.rs
```diff
@@ -1,46 +1,98 @@
...
+            // Not `::always()`, because actor can impose its own limits.
+            Interest::sometimes()
+        } else {
+            // Won't be ever allowed by the `.targets`.
+            Interest::never()
+        }
     }
 
     fn enabled(&self, meta: &Metadata<'_>, cx: Context<'_, S>) -> bool {
+        let config = self.inner.config.load();
```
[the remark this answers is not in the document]
GoldsteinE: Because config could change. `Cache` probably makes sense, yes.

examples/examples/usage/config.toml
```diff
@@ -26,6 +22,10 @@
 #path = "example.log"
 #format.with_location = false
 #format.with_module = false
+#
+# It's possible to override `max_level` for a specific target:
```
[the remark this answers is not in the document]
GoldsteinE: Fixed the wording.

PR #95, opened 2023-06-30
feat(init,configurer): allow to check configs without starting the system

laplab:


elfo-core/src/errors.rs
```diff
@@ -1,14 +1,36 @@
...
-    #[display(fmt = "invalid config")]
-    InvalidConfig,
+pub enum StartErrorReason {
+    #[display(fmt = "entrypoint failed: {}", _0)]
+    EntrypointFailed(#[error(not(source))] String),
+    #[display(fmt = "other: {}", _0)]
     Other(#[error(not(source))] String),
 }
 
+impl From<ConfigRejected> for StartErrorReason {
```
laplab, line 22: This is a bit awkward. Initially, I wanted to just add `#[derive(Display)]` to the `ConfigRejected` and call `to_string()` in order to extract the failure reason. However, this produces conflicting implementations of `From` trait, since we already have this in the codebase:

```rust
impl<R: Display> From<R> for ConfigRejected {
    fn from(reason: R) -> Self {
        Self {
            reason: reason.to_string(),
        }
    }
}
```

And `ConfigRejected` having `Display` trait implemented causes this to conflict with `impl From<T> for T` from Rust itself. Replacing `Display` with `ToString` above does not solve the problem, since `ToString` is implemented for `Display`.

Without negative trait bounds, there is not much of a choice left. Let me know if you have a workaround in mind.

elfo-core/src/init.rs
```diff
@@ -29,48 +29,50 @@ use crate::{
...
+            ) -> Result<()> {
+                match response {
+                    Ok(Ok(())) => Ok(()),
+                    Ok(Err(e)) => Err(StartError {
+                        group,
+                        reason: e.into(),
+                    }),
+                    Err(RequestError::Ignored) => Ok(()),
+                    Err(RequestError::Closed(..)) => Err(StartError {
+                        group,
```
laplab, line 49: We could include the whole `ActorGroup` in here, but I am not sure if this would be actually useful. There is no error recovery to be done, so the message is only for the end user to analyse.

elfo-core/src/init.rs
```diff
@@ -29,48 +29,50 @@ use crate::{
...
-        .map_err(|_| StartError::Other("entrypoint cannot start".into()))?;
+            let response = ctx
+                .request_to(group.addr, StartEntrypoint { is_check_only })
+                .resolve()
+                .await;
+            convert_response(group.name, response)
+        });
 
-    Ok(())
+    futures::future::try_join_all(futures).await.map(|_| ())
```
laplab, line 75: One thing I initially missed is that `try_join_all` will actually return only one error. So if there were multiple entrypoints and all of them failed, we will report only one of them, which might be unexpected. Maybe we should just wait for all of them? On the start of the system, all mailboxes should be empty, so I do not expect this to take long.

elfo-core/src/init.rs
```diff
@@ -29,48 +29,50 @@ use crate::{
...
-    } else {
-        Err(StartError::InvalidConfig)
-    }
-}
+                        ),
+                    }),
+                }
+            }
 
-async fn wait_entrypoints(ctx: &Context, topology: &Topology) -> Result<()> {
```
laplab, line 63: This is a proposal to remove `wait_entrypoints` step. Entrypoint is free to exit as soon as it finished doing its thing, so there can be a situation where `UpdateConfig` + `StartEntrypoint` are successful, but the `Ping` afterwards fails.

elfo/benches/round_trip.rs
```diff
@@ -153,7 +153,7 @@ async fn run<const FLAGS: Flags>(
         AnyConfig::default(),
     ));
 
-    elfo::_priv::do_start(topology, |ctx, _| async move {
+    elfo::_priv::do_start(topology, false, |ctx, _| async move {
```
laplab, line 156: Not a big fan of boolean flags in the middle of arguments, but this is a private interface anyway, so it might be okay. As an alternative, we can introduce enum for starting modes.

CHANGELOG.md
```diff
@@ -8,6 +8,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased] - ReleaseDate
 
+### Added
+- **BREAKING** init,configurer: allow to check configs without starting the system ([#95]).
```
loyd, line 12: Why `BREAKING`?

elfo-core/src/messages.rs
```diff
@@ -52,6 +52,28 @@ impl<R: Display> From<R> for ConfigRejected {
...
 }
 
+#[message(ret = Result<(), EntrypointFailed>)]
+#[derive(Constructor)]
+#[non_exhaustive]
+pub struct StartEntrypoint {
+    pub is_check_only: bool,
+}
+
+#[message]
```
loyd, line 62: It's better to use `message(part)` here to avoid adding to the message index. It's possible because `EntrypointFailed` isn't sent directly.

elfo/benches/round_trip.rs
loyd: flags are fine for private fns

CHANGELOG.md
loyd: The only thing is breaking here: moving `start` and `try_start` to the `init` submodule. It should be noted separately under the `Changed` section with the `BREAKING` disclaimer.

elfo-configurer/src/lib.rs
```diff
@@ -76,6 +76,34 @@ impl Configurer {
...
+                        }
+                        None
+                    }
+                    // In case the actor was restarted by the supervisor, no `StartEntrypoint`
+                    // message is sent and we need to process any incoming envelopes normally.
+                    e => Some(e),
+                })
+            }
+            Err(TryRecvError::Empty) => None,
+            Err(TryRecvError::Closed) => panic!("failed to recv first envelope"),
```
loyd, line 104: This group doesn't override the termination policy, so on `Terminate` just after restarting, it will get `Closed` here and panic. I prefer `return;` here

elfo-core/src/errors.rs
```diff
@@ -1,14 +1,40 @@
...
+#[non_exhaustive]
+pub struct EntrypointErrors {
+    pub errors: Vec<EntrypointError>,
+}
+
+impl Display for EntrypointErrors {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        writeln!(f, "entrypoint errors:")?;
+        for error in &self.errors {
+            writeln!(f, "- {}", error)?;
```
laplab, line 24: This way of listing errors does not compose. If this error is a part of a top-level list, no indentation is applied. An alternative is to list everything on the same line, which is not readable.

elfo-core/src/init.rs
```diff
@@ -29,50 +29,74 @@ use crate::{
...
+                    errors: vec![EntrypointError {
+                        group: group.name.clone(),
+                        reason: e.reason,
+                    }],
+                })),
+                Err(RequestError::Ignored) => Ok(()),
+                Err(RequestError::Closed(..)) => Err(StartError::Other(
+                    "config cannot be delivered to the entrypoint".into(),
+                )),
+            }?;
```
laplab, line 58: Just realized that we still can extract this into function, just pass the message for `StartError::Other` into it

elfo-core/src/errors.rs
```diff
@@ -1,14 +1,40 @@
...
 use derive_more::{Display, Error};
 
 #[derive(Debug, Display, Error)]
 #[non_exhaustive]
 pub enum StartError {
-    /// Configs must be valid at the start-up.
-    #[display(fmt = "invalid config")]
-    InvalidConfig,
+    #[display(fmt = "error starting entrypoints: {}", _0)]
+    EntrypointErrors(#[error(not(source))] EntrypointErrors),
```
loyd, line 9: Double `Error` in `StartError::EntrypointErrors`. What about `EntrypointFailed` or simply `Entrypoint`?

elfo-core/src/init.rs
```diff
@@ -85,7 +109,20 @@ pub async fn start(topology: Topology) {
...
+    res
+}
+
+/// Starts node in "check only" mode. Entrypoints are started, then the system
+/// is immediately gracefully terminated.
+pub async fn check_only(topology: Topology) -> Result<()> {
+    let res = do_start(topology, true, do_termination).await;
 
     if res.is_err() {
         // XXX: give enough time to the logger.
```
loyd, line 128: The logger won't be started in the only_check mode, so you don't need to do this hack here.

elfo-core/src/messages.rs
```diff
@@ -52,6 +52,20 @@ impl<R: Display> From<R> for ConfigRejected {
...
+#[non_exhaustive]
+pub struct StartEntrypoint {
+    pub is_check_only: bool,
+}
+
+#[message(part)]
+#[derive(Constructor)]
+#[non_exhaustive]
+pub struct EntrypointFailed {
+    pub reason: String,
```
loyd, line 66: `errors: Vec<_>`?

elfo-core/src/messages.rs
```diff
@@ -52,6 +52,20 @@ impl<R: Display> From<R> for ConfigRejected {
     }
 }
 
+#[message(ret = Result<(), EntrypointFailed>)]
+#[derive(Constructor)]
+#[non_exhaustive]
+pub struct StartEntrypoint {
```
loyd, line 58: I think we have to stick to the same style in elfo. E.g. in the configurer's protocol: `ReloadConfigs` and `ReloadConfigsRejected`.

elfo-core/src/errors.rs
loyd: On the same line is ok for the default style (but using `,` or `;` as a separator). It can be altered in the future by using the alternate mode of `Display` (`#`). Anyway, users can implement any useful form for them.

Alternatively, we can inspire by https://docs.rs/eyre/latest/eyre/struct.Report.html#display-representations

elfo-configurer/src/lib.rs
```diff
@@ -101,10 +136,21 @@ impl Configurer {
         }
     }
 
-    async fn load_and_update_configs(
-        &mut self,
-        force: bool,
-    ) -> Result<(), Vec<ReloadConfigsError>> {
+    async fn check_configs(&self) -> Result<(), Vec<ReloadConfigsError>> {
```
loyd, line 139: It should be `load_and_check_configs` in order to be consistent with `load_and_update_configs`.

elfo-configurer/src/lib.rs
```diff
@@ -76,6 +76,38 @@ impl Configurer {
...
+                msg!(match e {
+                    // We do not expect the first message to be `ConfigUpdated` because when the
+                    // actor is first started, `UpdateConfig` is consumed by the supervisor.
+                    (StartEntrypoint { is_check_only, .. }, token) => {
+                        if is_check_only {
+                            match self.check_configs().await {
+                                Ok(_) => self.ctx.respond(token, Ok(())),
+                                Err(errors) => {
+                                    // TODO laplab: this is slow.
+                                    let mut reason = String::from("config check failed:\n");
```
loyd, line 90: Just replace `reason` with a vector of errors.

elfo-configurer/src/lib.rs
```diff
@@ -76,18 +79,68 @@ impl Configurer {
...
+
+                        if is_check_only {
+                            match self.load_and_check_configs().await {
+                                Ok(_) => self.ctx.respond(token, Ok(())),
+                                Err(errors) => self.ctx.respond(
+                                    token,
+                                    Err(StartEntrypointRejected::new(convert_to_protocol(errors))),
+                                ),
+                            }
+                            return;
```
laplab, line 106: This should happen in both check and non-check mode actually.

elfo-core/src/init.rs
```diff
@@ -29,50 +29,73 @@ use crate::{
...
+                        .errors
+                        .into_iter()
+                        .map(|e| GroupError {
+                            group: e.group,
+                            reason: e.reason,
+                        })
+                        .collect();
+                    group_errors.push(GroupError {
+                        group: Some(group.name),
+                        reason: "entrypoint failed to start".into(),
```
loyd, line 73: It's misleading. I believe only original errors should be shown.

elfo-configurer/src/protocol.rs
```diff
@@ -29,6 +29,6 @@ pub struct ReloadConfigsRejected {
 #[message(part)]
 #[non_exhaustive]
 pub struct ReloadConfigsError {
-    // TODO: meta
+    pub group: Option<String>,
```
loyd, line 32: Should it be mandatory with "configurers" (actually the real name according to topology) if the config file is invalid?

elfo-configurer/src/lib.rs
```diff
@@ -117,33 +172,55 @@ impl Configurer {
...
+                reason: error.to_string(),
+            }]
+        })
+    }
+
+    async fn load_and_check_configs(&self) -> Result<(), Vec<ReloadConfigsError>> {
+        let configs = self.load_configs().await?;
+
+        let system_configs = match_configs(&self.topology, &configs, TopologyFilter::System);
+        self.validate_all(&system_configs).await?;
```
loyd, line 198: Maybe check against both system and user groups to collect all possible errors at once (`TopologyFilter::All`)?

elfo-core/src/errors.rs
```diff
@@ -1,12 +1,103 @@
...
+                i += 1;
+            }
+        }
+        Ok(())
+    }
+}
+
+#[derive(Clone, Debug, Error)]
+#[non_exhaustive]
+pub struct GroupError {
```
loyd, line 89: `StartGroupError`

loyd (approved):

PR #99, opened 2023-07-07
feat(network): lz4 compression

laplab:


elfo-network/src/frame/read.rs
```diff
@@ -0,0 +1,224 @@
...
+                                // If there is still not enough space after the shift, we need to
+                                // reallocate.
+                                let new_buffer_size = self.filled + rounded_length;
+                                self.compressed_buffer
+                                    .resize(new_buffer_size, MaybeUninit::uninit());
+                            }
+                        }
+
+                        let compressed_slice = unsafe {
+                            slice_assume_init_mut(&mut self.compressed_buffer[self.filled..])
```
laplab, line 166: I am pretty sure this is UB. We are guaranteed to have uninit data here. I would return `&mut [MaybeUninit<u8>]` here, but the `OwnedReadHalf::read` accepts only `&mut [u8]`

elfo-network/src/frame/lz4.rs
```diff
@@ -0,0 +1,138 @@
...
+    ) -> Result<DecodeState<&[u8]>> {
+        if raw.len() < 4 {
+            return Ok(DecodeState::NeedMoreData {
+                total_length_estimate: 4,
+            });
+        }
+
+        let mut input = Cursor::new(raw);
+        let frame_size = input.read_u32::<LittleEndian>()? as usize;
+        if frame_size >= 200_000_000 {
```
loyd, line 68: Firstly, this constant appears twice

Secondly, this constant doesn't match the envelope size.

I agree that some limits should be used, but the current implementation is clumsy. Let's move to a constant and write `TODO` for now

elfo-network/src/frame/lz4.rs
```diff
@@ -0,0 +1,138 @@
...
+                total_length_estimate: frame_size,
+            });
+        }
+
+        let decompressed_size = input.read_u32::<LittleEndian>()? as usize;
+        if decompressed_size >= 200_000_000 {
+            return Err(eyre!("decompressed size is too big"));
+        }
+
+        self.buffer.clear();
```
loyd, line 83: We're losing 1/3 of LZ4's performance here. I think we should simply remove this line.

Ideally, the next line should be prepended with `if self.buffer.len() < decompressed_size` to avoid extra (nonsimdable) filling.

elfo-network/src/frame/lz4.rs
```diff
@@ -0,0 +1,138 @@
...
+        if actual_size != decompressed_size {
+            return Err(eyre!(
+                "expected to decompress {} bytes, got {}",
+                decompressed_size,
+                actual_size
+            ));
+        }
+
+        stats.total_compressed_bytes += frame_size as u64;
+        stats.total_uncompressed_bytes += decompressed_size as u64;
```
loyd, line 98: According to the naming (compressed, uncompressed), it can be considered correct.
However, we're actually interested in measuring the real traffic, so `total_compressed_bytes` must be increased before any possible errors.

elfo-network/src/frame/lz4.rs
```diff
@@ -0,0 +1,138 @@
...
+
+use byteorder::{LittleEndian, ReadBytesExt, WriteBytesExt};
+use std::io::Cursor;
+
+use eyre::{eyre, Result};
+
+use crate::codec::decode::DecodeState;
+
+pub(crate) struct LZ4Buffer {
+    buffer: Vec<u8>,
```
loyd, line 29: This field changes semantics depending on the usage: it stores compressed data in one case and uncompressed in another, it should be commented.

elfo-network/src/frame/read.rs
```diff
@@ -0,0 +1,235 @@
...
+    frame::{
+        buffers::{ReadBuffer, COMPRESSED_DATA_BUFFER_CAPACITY, DECOMPRESSED_DATA_BUFFER_CAPACITY},
+        lz4::{DecompressStats, LZ4Buffer},
+    },
+};
+
+use derive_more::Display;
+use eyre::{eyre, Result};
+
+#[derive(Display)]
```
loyd, line 16: It's strange to implement `Display` for complex structures with logic, prefer printing `CompressionAlgorithm` instead

elfo-network/src/frame/buffers.rs
```diff
@@ -0,0 +1,74 @@
...
+
+    pub(crate) fn len(&self) -> usize {
+        self.filled - self.position
+    }
+
+    pub(crate) fn extend_to_contain(&mut self, new_length: usize) {
+        // We round up the number of requested bytes to be a multiple of the chunk size
+        // to avoid reading too few bytes per syscall.
+        let current_length = self.len();
+        debug_assert!(new_length > current_length);
```
loyd, line 47: I don't understand this assertion. `new_length` can easily be `4` (returned by the decoder), but the buffer can contain a lot of already-used bytes.

elfo-network/src/frame/read.rs
```diff
@@ -0,0 +1,235 @@
...
+                })
+            }
+            DecodeState::Done {
+                bytes_consumed,
+                decoded,
+            } => {
+                self.stats.decompress_stats.total_uncompressed_bytes += bytes_consumed as u64;
+                self.buffer.consume(bytes_consumed);
+                Ok(FramedReadState::Done {
+                    decoded: Some(decoded),
```
loyd, line 222: It's suspicious that there is no `None` variant, it indicates for me not ideal LZ4 implementation.

loyd (changes requested):

elfo-network/src/socket.rs
```diff
@@ -1,145 +1,433 @@
...
 
+        buf.put_u64_le(HANDSHAKE_MAGIC);
+        buf.put_u8(self.version);
+        buf.put_u16_le(self.node_no);
+        buf.put_u64_le(self.launch_id.into());
+        buf.put_u32_le(self.capabilities.bits());
+        buf.put_slice(&[0; HANDSHAKE_RESERVED_LENGTH]);
+
+        let result = buf.freeze();
+        debug_assert!(result.len() == HANDSHAKE_LENGTH);
```
loyd, line 78: `_eq` to see much helpful error

loyd (approved):

PR #101, opened 2023-07-25
fix(network): notify remote actors if request parsing failed

laplab:


elfo-network/src/connection.rs
```diff
@@ -331,12 +332,44 @@ struct SocketReader {
...
+                        .get(self.handle_addr)
+                        .expect("bug: remote group is missing in the address book");
+                    let token = ResponseToken::new(
+                        details.sender,
+                        details.request_id,
+                        details.trace_id,
+                        self.ctx.book().clone(),
+                    );
+                    sender.respond(token, Err(RequestError::Failed));
+                    error!(
```
loyd, line 360: The same event produces the two related errors (in the decoder and here). I think it's better to provide more information in the first place and, optionally, `debug!` here.

I think later, `Skipped` and `RequestSkipped` will be merged anyway (because we should also handle `Skipped` to fix flow control state), so for now it's ok just to replace this level with `debug` one

loyd (changes requested):

elfo-network/src/frame/read.rs
```diff
@@ -31,6 +31,10 @@ pub(crate) enum FramedReadState<'a> {
     /// The stategy needs more data written at the beginning of the specified
     /// `buffer`.
     NeedMoreData { buffer: &'a mut [u8] },
+    /// The strategy failed to read a request message. This requires special
```
loyd, line 34: Misleading comment

elfo-network/src/connection.rs
```diff
@@ -361,6 +378,49 @@ impl SocketReader {
...
+            }
+        };
+
+        if let Some(envelope) = update_flow.map(make_system_envelope) {
+            let item = KanalItem::simple(Addr::NULL, envelope);
+            self.tx.try_send(item).unwrap();
+        }
+
+        if details.kind == KIND_REQUEST_ALL || details.kind == KIND_REQUEST_ANY {
+            scope::set_trace_id(details.trace_id);
```
loyd, line 404: It should be the first statement in the method. Even better to move above
```
self.handle_skipped_envelope(details);
```

elfo-network/src/connection.rs
```diff
@@ -361,6 +378,49 @@ impl SocketReader {
         ConnectionClosed
     }
 
+    /// Ensures that messages that were skipped due to errors during decoding
+    /// are properly accounted for in flow control. Also notifies the remote
+    /// actor if the message was a request in order to avoid indefinite
+    /// waiting from the remote actor's side.
+    fn handle_skipped_envelope(&self, details: EnvelopeDetails) {
```
loyd, line 385: `handle_skipped_message` (or rename other `handle_*` messages)

loyd (approved):

PR #106, opened 2023-08-23
fix(logger): properly filter tracing-log entries

GoldsteinE:


elfo-logger/src/lib.rs
```diff
@@ -82,6 +82,19 @@ pub fn init() -> Blueprint {
...
+    #[cfg(feature = "tracing-log")]
+    {
+        if let Err(e) = tracing_log::LogTracer::init() {
+            tracing::error!(
+                // REVIEW: elfo doesn't have magic Value as whizz does, so just allocating here
+                error = &*e.to_string(),
+                "can't integrate with log adapter, logs produced via `log` crate might be not available",
+            );
+        } else {
+            log::info!("Initialized tracing_log adapter");
```
loyd, line 94: all logs in elfo starts with lowercase

elfo-logger/src/filtering_layer.rs
```diff
@@ -76,6 +79,32 @@ impl<S: Subscriber> Layer<S> for FilteringLayer {
...
+        if let Some(&log_name) = self.inner.log_metadata_name.get() {
+            // We've already got logs from `tracing-log`, just compare str
+            // pointers for performance.
+            if std::ptr::eq(log_name, name) {
+                let config = self.inner.config.load();
+                if !config.targets.would_enable(meta.target(), meta.level()) {
+                    return false;
+                }
+            }
+        } else if name == "log record" {
```
loyd, line 95: If the `tracing-log` feature is disabled, we shouldn't check and even use `log_metadata_name` at all.

loyd (approved):

PR #110, opened 2023-09-12
fix(core): forget tokens in duplicated envelopes

laplab:


elfo/tests/request_routing.rs
```diff
@@ -0,0 +1,103 @@
...
+            panic!("thief should not be started");
+        });
+
+    let _ = tracing_subscriber::fmt()
+        .with_target(false)
+        .with_env_filter(tracing_subscriber::EnvFilter::from_default_env())
+        .with_test_writer()
+        .try_init();
+
+    let config = toml::toml! {
```
loyd, line 71: empty config should work too

loyd (approved):

PR #116, opened 2023-11-01
fix(network): unrouted network requests on the remote node don't lead to the caller getting stuck

sargarass:


loyd (approved):

elfo-core/src/topology.rs
```diff
@@ -470,11 +473,14 @@ cfg_network!({
     // Nothing for now, reserved for future use.
     pub struct NodeDiscovery(());
 
+    // TODO: should undo register_remote in drop
```
loyd, line 476: ?

PR #118, opened 2023-12-14
feat(core): the restarting policy can be overridden via configuration for each actor.

sargarass:
`RestartParams` have been added for `RestartPolicy::always(..)` and `RestartPolicy::on_failure(..)`. 
The linear backoff has been replaced with an exponential approach, with a configurable limit for restarts.

BREAKING CHANGE: The default RestartPolicy is set to `RestartPolicy::never()`, `RestartPolicy::always` and `RestartPolicy::on_failure` have args .


elfo-core/src/supervisor/backoff.rs
```diff
@@ -1,20 +1,20 @@
...
+    next_backoff: Option<Duration>,
     start_time: Instant,
+    restart_count: u64,
 }
 
 impl Default for Backoff {
     fn default() -> Self {
         Self {
-            next_backoff: BACKOFF_STEP,
+            next_backoff: None,
```
loyd, line 15: It seems it's the only place where it can become `None` =(

elfo-core/src/supervisor.rs
```diff
@@ -323,26 +332,33 @@ where
...
+            actor.set_status(new_status);
+
+            drop(object);
+
+            let restart_after = restarting_allowed
+                .then(|| {
+                    let restart_params = restart_policy
+                        .restart_params()
+                        .expect("restart params are set if actor can restart");
+                    backoff.next(&restart_params)
```
loyd, line 357: It means that until an actor dies, his policy cannot be changed. So, it's better to calculate the actual restart policy here, not at the start of `fn spawn`

elfo-core/src/supervisor/backoff.rs
loyd: I think it's better to remove `next_backoff` totally (his function is duplicated by `restart_count`)

loyd (changes requested):

elfo-core/src/supervisor/backoff.rs
```diff
@@ -1,20 +1,20 @@
 use std::time::Duration;
 
+use crate::RestartParams;
 use elfo_utils::time::Instant;
 
-const BACKOFF_STEP: Duration = Duration::from_secs(5);
-const MAX_BACKOFF: Duration = Duration::from_secs(30);
-
 pub(crate) struct Backoff {
```
loyd, line 6: It's not obvious to me now why this struct is placed here, not inside `mod restarting`. Or the module should be named `restart_policy`.

elfo-core/src/supervisor.rs
```diff
@@ -323,26 +332,33 @@ where
...
-            let need_to_restart = should_restart && !sv.control.read().stop_spawning;
-            if need_to_restart {
-                let after = backoff.next();
-
+            let object = sv.objects.get(&key).expect("where is the current actor?");
+            let actor = object.as_actor().expect("a supervisor stores only actors");
+
+            let rp_override = actor.restart_policy();
+            // Select the restart policy with the next priority:
+            //     ctx.set_restart_policy(..) > config > blueprint (sv.restart_policy).
```
loyd, line 340: > sv.restart_policy

I think it should be named as `default_restart_policy` now to avoid confusion for readers

elfo-core/src/supervisor/backoff.rs
```diff
@@ -24,53 +24,125 @@ impl Backoff {
...
+            Some(params.max_backoff)
+        } else {
+            let nanos = current_nanos * params.factor;
+            Some(nanos_to_duration(nanos).max(params.min_backoff))
+        };
+        Some(current)
     }
 }
 
+fn duration_to_nanos(d: Duration) -> f64 {
```
loyd, line 60: It has the same precision as simple `d.as_secs_f64()`

elfo-core/src/restarting/restart_policy.rs
```diff
@@ -0,0 +1,115 @@
...
+    ///
+    /// `min_backoff` is used by default.
+    pub fn auto_reset(self, auto_reset: Duration) -> Self {
+        Self { auto_reset, ..self }
+    }
+
+    /// Sets the factor used to calculate the next backoff duration.
+    ///
+    /// Factor of `2.0` is used by default.
+    pub fn factor(self, factor: f32) -> Self {
```
loyd, line 99: f32 here is also questionable

elfo-core/src/supervisor/backoff.rs
sargarass: moved to the mod and changed the struct name to `RestartBackoff`

elfo-core/src/restarting/restart_policy.rs
```diff
@@ -0,0 +1,115 @@
...
+
+    /// The duration considered sufficiently long-running for an actor. After
+    /// this, the backoff strategy automatically resets, including retry
+    /// counting (the next attempt becomes the first retry).
+    /// Therefore, setting the `auto_reset` to small values, such as
+    /// `Duration::ZERO`, can result in the absence of a limit on the maximum
+    /// number of retries.
+    ///
+    /// `min_backoff` is used by default.
+    pub fn auto_reset(self, auto_reset: Duration) -> Self {
```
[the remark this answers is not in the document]
sargarass: used the suggested pattern.

elfo-core/src/context.rs
```diff
@@ -624,6 +624,41 @@ impl<C, K> Context<C, K> {
...
+    ///     }
+    ///     ActorStartCause::Restarted => {
+    ///         // The actor started due to the restart policy.
+    ///     }
+    ///     _ => {}
+    /// }
+    /// # }
+    /// ```
+    pub fn start_info(&self) -> ActorStartInfo {
+        self.actor
```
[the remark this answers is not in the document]
sargarass: moved to Context

elfo-core/src/supervisor.rs
sargarass: It will be unexpected if I explicitly change the `restart_policy` via the `ctx` and regain it back when the configuration is reloaded.

elfo-core/src/config.rs
```diff
@@ -162,6 +162,7 @@ pub(crate) struct SystemConfig {
     pub(crate) logging: crate::logging::LoggingConfig,
     pub(crate) dumping: crate::dumping::DumpingConfig,
     pub(crate) telemetry: crate::telemetry::TelemetryConfig,
+    pub(crate) restarting: crate::restarting::RestartingConfig,
```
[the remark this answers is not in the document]
sargarass: ok

elfo-core/src/supervisor.rs
sargarass: I renamed `sv.restart_policy` to `default_restart_policy` and introduced a new attribute, `group_restart_policy`, within `sv.control`. This attribute is updated during a `config_update` if specified in the configuration. If not explicitly defined, it defaults back to `default_restart_policy`. 

So in the code above we will chose either the actor `rp_override` or the `group_restart_policy`

elfo-core/src/supervisor.rs
sargarass: true, made adjustments

elfo-core/src/supervisor/backoff.rs
sargarass: I simplified a bit backoff.rs

elfo-core/src/restarting/config.rs
```diff
@@ -0,0 +1,48 @@
...
+struct RestartParamsConfig {
+    #[serde(with = "humantime_serde")]
+    min_backoff: Duration,
+    #[serde(with = "humantime_serde")]
+    max_backoff: Duration,
+    #[serde(with = "humantime_serde", default)]
+    auto_reset: Option<Duration>,
+    max_retries: Option<NonZeroU64>,
+    #[serde(default)]
+    factor: Option<f64>,
```
loyd, line 28: If defined, it MUST be `> 0.0`

elfo-core/src/restarting/config.rs
```diff
@@ -0,0 +1,48 @@
...
+#[derive(Debug, Clone, Deserialize)]
+struct RestartParamsConfig {
+    #[serde(with = "humantime_serde")]
+    min_backoff: Duration,
+    #[serde(with = "humantime_serde")]
+    max_backoff: Duration,
+    #[serde(with = "humantime_serde", default)]
+    auto_reset: Option<Duration>,
+    max_retries: Option<NonZeroU64>,
+    #[serde(default)]
```
loyd, line 27: What's the purpose of this annotation?

sargarass: removed

elfo-core/src/restarting/config.rs
sargarass: will emit warning and default to safe value to not panic at runtime.

elfo-core/src/restarting/backoff.rs
```diff
@@ -0,0 +1,135 @@
...
+        let delay = (params.min_backoff.as_secs_f64() * params.factor.powf(self.power as f64))
+            .clamp(
+                params.min_backoff.as_secs_f64(),
+                params.max_backoff.as_secs_f64(),
+            );
+
+        self.power += 1;
+        // Check for overflow, if overflow is detected set the current delay to maximal
+        let delay = if delay.is_finite() {
+            Duration::from_secs_f64(delay)
```
loyd, line 49: Add a comment that factor is checked against negativity and therefore it doesn't lead to panic

loyd (approved):

PR #119, opened 2024-01-03
refactor(core): dump AnyMessage as a struct

GoldsteinE:
Any other serialization mode still uses a tuple representation for compactness. Using struct in the dumping mode makes it more readable.

### DIFF
--- a/elfo-core/src/message.rs
+++ b/elfo-core/src/message.rs
@@ -6,14 +6,14 @@ use metrics::Label;
 use once_cell::sync::Lazy;
 use serde::{
     de::{DeserializeSeed, SeqAccess, Visitor},
-    ser::SerializeTuple,
+    ser::{SerializeStruct as _, SerializeTuple as _},
     Deserialize, Deserializer, Serialize,
 };
 use smallbox::{smallbox, SmallBox};
 
 use elfo_utils::unlikely;
 
-use crate::dumping;
+use crate::{dumping, scope::SerdeMode};
 
 pub trait Message: fmt::Debug + Clone + Any + Send + Serialize + for<'de> Deserialize<'de> {
     #[inline(always)]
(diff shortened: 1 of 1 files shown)


elfo-core/src/message.rs
```diff
@@ -150,15 +150,21 @@ impl Serialize for AnyMessage {
...
-        tuple.serialize_element(self.protocol())?;
-        tuple.serialize_element(self.name())?;
-
         // TODO: avoid allocation here
         let erased_msg = self._erase();
-        tuple.serialize_element(&*erased_msg)?;
-
-        tuple.end()
+        if crate::scope::serde_mode() == SerdeMode::Dumping {
+            let mut fields = serializer.serialize_struct("RobotMessage", 3)?;
```
jazvit, line 156: Message/AnyMessage

jazvit (approved):
LGTM

elfo-core/src/message.rs
```diff
@@ -150,15 +150,21 @@ impl Serialize for AnyMessage {
...
-        let mut tuple = serializer.serialize_tuple(3)?;
-        tuple.serialize_element(self.protocol())?;
-        tuple.serialize_element(self.name())?;
-
         // TODO: avoid allocation here
         let erased_msg = self._erase();
-        tuple.serialize_element(&*erased_msg)?;
-
-        tuple.end()
+        if crate::scope::serde_mode() == SerdeMode::Dumping {
```
loyd, line 155: We should comment that `Serialize` and `Deserialize` instances run only if `AnyMessage` is used inside other messages (although it's unstable anyway).

elfo-core/src/message.rs
loyd: `AnyMessage`. Messages are encoded differently.

loyd (approved):

PR #120, opened 2024-01-30
feat(core): use available memory as a trigger for OOM shutdown

White-Oak:
`avail < total * (1 - ratio)` threshold is now used instead of old `used > total * ratio` to shutdown a service when tracking memory usage. 
Additionally, total memory in bytes, and used/available memory as percentages of total memory are now being logged on OOM.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - core: changed OOM prevention trigger (that gracefully shutdowns the service) from old `used > total * ratio` to new `avail < total * (1 - ratio)`.  ([#120])
 
+[#120] - https://github.com/elfo-rs/elfo/pull/120
+
 ## [0.2.0-alpha.12] - 2023-12-21
 ### Changed
 - **BREAKING** restarting: the default restart policy is `RestartPolicy::never()` now ([#118]).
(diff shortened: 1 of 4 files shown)


elfo-core/src/memory_tracker.rs
```diff
@@ -41,21 +58,24 @@ mod proc_stats {
...
     pub(super) fn get() -> Result<MemoryStats, String> {
         const PROC_SELF_STATM: &str = "/proc/self/statm";
-        const PROC_MEMINFO: &str = "/proc/meminfo";
         const PAGE_SIZE: usize = 4096; // TODO: use `sysconf(_SC_PAGESIZE)`
 
-        // Get `total`.
+        // Assumes `MemTotal` is always the 1st line in `/proc/meminfo`.
+        const MEM_TOTAL_LINE_NO: usize = 0;
+        // Assumes `MemAvailable` is always the 3rd line in `/proc/meminfo`.
+        const MEM_AVAILABLE_LINE_NO: usize = 2;
```
loyd, line 70: I don't think we should rely on the exact position in the file.

loyd (changes requested):

elfo-core/src/memory_tracker.rs
White-Oak: Noted, changed

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 <!-- next-header -->
 
 ## [Unreleased] - ReleaseDate
+### Changed
+- core: changed OOM prevention trigger (that gracefully shutdowns the service) from old `used > total * ratio` to new `avail < total * (1 - ratio)`.  ([#120])
```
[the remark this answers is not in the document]
White-Oak: Thanks! Missed that part by mistake, added

loyd (approved):

PR #121, opened 2024-02-11
Test utils for extracting messages and requests from Envelope

RicoGit:
HI! Added convenient methods for extracting specific message and requests type from Envelope in tests.

### DIFF
--- a/elfo-core/src/init.rs
+++ b/elfo-core/src/init.rs
@@ -10,7 +10,10 @@ use tracing::{error, info, level_filters::LevelFilter, warn};
 use elfo_utils::time::Instant;
 
 #[cfg(target_os = "linux")]
-use crate::{memory_tracker::MemoryTracker, time::Interval};
+use crate::{
+    memory_tracker::{MemoryCheckResult, MemoryTracker},
+    time::Interval,
+};
 
 use crate::{
     actor::{Actor, ActorMeta, ActorStartInfo, ActorStatus},
@@ -19,7 +22,6 @@ use crate::{
     context::Context,
     demux::Demux,
     errors::{RequestError, StartError, StartGroupError},
-    memory_tracker::MemoryCheckResult,
     message,
     messages::{StartEntrypoint, Terminate, UpdateConfig},
     object::Object,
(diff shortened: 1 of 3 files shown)


elfo-core/src/init.rs
```diff
@@ -19,7 +19,6 @@ use crate::{
     context::Context,
     demux::Demux,
     errors::{RequestError, StartError, StartGroupError},
-    memory_tracker::MemoryCheckResult,
```
RicoGit, line 22: This import fails compilation for non-linux OS. Moved it below under `#[cfg(target_os = "linux")]` section.

elfo-core/src/request_table.rs
```diff
@@ -303,6 +303,11 @@ impl<R> ResponseToken<R> {
     pub fn is_forgotten(&self) -> bool {
         self.data.is_none()
     }
+
+    #[cfg(any(test, feature = "test-util"))]
+    pub fn test() -> Self {
+        ResponseToken::forgotten()
+    }
```
RicoGit, line 310: Didn't find a way to create `ResponseToken` in test without this change.

loyd: mark `forgotten()` as `#[doc(hidden)] #[inline]` and use it for tests inside elfo.

the test-util feature is supposed to be used in user code, where response tokens should never be possible to create

elfo-test/src/util.rs
```diff
@@ -0,0 +1,73 @@
...
+            elfo_core::dumping::extract_name_by_type::<M>(),
+            msg
+        ),
+    })
+}
+
+/// Extracts request message with type [`R`] from [`Envelope`], panics
+/// otherwise.
+#[track_caller]
+pub fn extract_req<R: Request>(msg: Envelope) -> (R, ResponseToken<R>) {
```
loyd, line 19: `extract_request`

elfo-test/src/util.rs
```diff
@@ -0,0 +1,73 @@
...
+                Envelope::new(TestMessage, MessageKind::Regular { sender: Addr::NULL }).upcast();
+            assert_eq!(extract_msg::<TestMessage>(envelop), TestMessage);
+        });
+    }
+
+    #[test]
+    fn extract_request() {
+        create_scope().sync_within(|| {
+            let envelop =
+                Envelope::new(TestRequest, MessageKind::RequestAll(ResponseToken::test())).upcast();
```
loyd, line 57: It's possible to use here `MessageKind::Regular` to avoid modification of `ResponseToken` API

Issue #125, opened 2024-05-03
Panic at config deserialization on startup makes elfo stuck


Arjentix:
Example:

```rust
use elfo::{batteries, Topology};

#[tokio::main]
async fn main() {
    let topology = Topology::empty();

    let configurers = topology.local("system.configurers").entrypoint();
    configurers.mount(batteries::configurer::from_path(&topology, "config.toml"));

    let stub_group = topology.local("stub");
    stub_group.mount(stub::new());

    elfo::init::start(topology).await;
}

mod stub {
    use elfo::{ActorGroup, Blueprint};

    pub fn new() -> Blueprint {
        ActorGroup::new().config::<Config>().exec(|_ctx| async {
            println!("Stub");
        })
    }

    #[derive(Debug)]
    struct Config;

    impl<'de> serde::de::Deserialize<'de> for Config {
        fn deserialize<D>(_deserializer: D) -> Result<Self, D::Error>
        where
            D: serde::Deserializer<'de>,
        {
            println!("Deserializing");
            unimplemented!()
        }
    }
}

```

See full example here: https://github.com/Arjentix/elfo-bug

PR #126, opened 2024-05-03
fix(core): do not ignore error while sending starting message

Arjentix:
Thanks to @sargarass.

This fix passes my test: https://github.com/Arjentix/elfo-bug

Closes: #125


loyd:
`RequestError` will be revised anyway, so now it's ok to react to them similarly.

LGTM

PR #127, opened 2024-05-25
Implement intrusive mailbox

### LINKED ISSUE #52: Revise the mailbox implementation
Some variants:
elfo-rfc-mailbox drawio

loyd:
It's better to start the review with `mailbox.rs`, then `message/repr.rs`, `message/any.rs`, and `envelope.rs`.

Closes #52


elfo-core/src/mailbox.rs
```diff
@@ -1,56 +1,166 @@
...
+            if let Some(envelope) = self.queue.dequeue() {
+                self.tx_semaphore.add_permits(1);
+                return RecvResult::Data(envelope);
+            }
+
+            if self.tx_semaphore.is_closed() {
+                return self.on_close();
+            }
+
+            self.rx_notify.notified().await;
```
GoldsteinE, line 152: I’m not sure this is correct in presence of multiple concurrent consumers (which are currently possible per the comment above).

Consider this timeline (queue is empty at the start):
**Receiver 1:** tries and fails to `.dequeue()`
**Receiver 2:** tries and fails to `.dequeue()`
**Sender:** queues a message
**Sender:** queues another message
**Receiver 1:** reaches `.notified()`, which consumes the notification and returns
**Receiver 1:** takes one message from the queue, leaving one queued
**Receiver 2:** reaches `.notified()` and hangs, despite there being a message in the queue

I’m not sure whether this case is something we’re concerned about, but it seems possible.

elfo-core/src/message/repr.rs
```diff
@@ -0,0 +1,340 @@
...
+/// using the virtual table ([`MessageVTable`]) for all operations.
+///
+/// [`AnyMessage`]: crate::AnyMessage
+/// [`Envelope`]: crate::Envelope
+// `vtable` must be first
+#[repr(C)]
+// It's `pub` only because used in private methods of `AnyMessage`.
+// Actually, it's not reexported at all.
+#[doc(hidden)]
+pub struct MessageRepr<M = ()> {
```
GoldsteinE, line 73: If I understood correctly, `M = ()` stands for “any message”, kind of like `void*`. Maybe a type with more descriptive name could be used here instead of `()`?

elfo-core/src/message/repr.rs
```diff
@@ -0,0 +1,340 @@
...
+    pub(super) labels: [Label; 2],    // protocol + name for `metrics`
+    pub(super) dumping_allowed: bool, // TODO: introduce `DumpingMode`.
+    #[cfg(feature = "network")]
+    pub(super) read_msgpack: unsafe fn(&[u8], NonNull<MessageRepr>) -> Result<(), decode::Error>,
+    #[cfg(feature = "network")]
+    #[allow(clippy::type_complexity)]
+    pub(super) write_msgpack:
+        unsafe fn(NonNull<MessageRepr>, &mut Vec<u8>, usize) -> Result<(), encode::Error>,
+    pub(super) debug: unsafe fn(NonNull<MessageRepr>, &mut fmt::Formatter<'_>) -> fmt::Result,
+    pub(super) clone: unsafe fn(NonNull<MessageRepr>, NonNull<MessageRepr>),
```
GoldsteinE, line 125: I recently learned that you can actually annotate names in function pointer types, like

```suggestion
    pub(super) clone: unsafe fn(src: NonNull<MessageRepr>, dest: NonNull<MessageRepr>),
```

which might be a bit more descriptive, given that most arguments here have the same types.

elfo-core/src/message/repr.rs
```diff
@@ -0,0 +1,340 @@
...
+
+/// Generic vtable's functions for monomorphization in [`MessageVTable::new()`].
+///
+/// All functions are `unsafe` because they work with raw pointers.
+///
+/// # Safety
+///
+/// Common safety requirements for all functions:
+/// * input pointers (`ptr`) must be [valid] for reading `MessageRepr<M>`.
+/// * output pointers (`out_ptr`) must be [valid] for writing `MessageRepr<M>`.
```
GoldsteinE, line 178: Output pointers also should not contain any previous values, given that `ptr::write()` will discard them without running destructors.

elfo-core/src/message/repr.rs
```diff
@@ -0,0 +1,340 @@
...
+            .map(|vtable| ((vtable.protocol, vtable.name), *vtable))
+            .collect()
+    });
+
+impl MessageVTable {
+    /// Finds a vtable by protocol and name.
+    /// Used for deserialization of `AnyMessage` and in networking.
+    pub(crate) fn lookup(protocol: &str, name: &str) -> Option<&'static Self> {
+        // Extend lifetimes to static in order to get `(&'static str, &'static str)`.
+        // SAFETY: this pair doesn't overlive the function.
```
GoldsteinE, line 285: That’s mildly scary. Using a custom key type allows to avoid any unsafety here: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=04a560d64fba49b3a3a467ea6dd673e4

elfo-core/src/message/repr.rs
```diff
@@ -0,0 +1,340 @@
...
+    use std::io;
+
+    struct LimitedWrite<W>(W, usize);
+
+    impl<W: io::Write> io::Write for LimitedWrite<W> {
+        #[inline]
+        fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
+            if buf.len() > self.1 {
+                self.1 = 0;
+                return Ok(0);
```
GoldsteinE, line 328: Is there a test for over-the-limit `.write_msgpack()`? I think it’s worthwhile to check that rmp-serde handles `Ok(0)` correctly.

elfo-core/src/message/repr.rs
GoldsteinE: I wonder if maybe having a separate type for an “uninit `MessageRepr`” would make more sense here (so `alloc_repr()` would return `UninitMessageRepr`, and e.g. `clone()` would accept one as `out_ptr`).

elfo-core/src/message.rs
```diff
@@ -26,439 +33,126 @@ pub trait Message: fmt::Debug + Clone + Any + Send + Serialize + for<'de> Deseri
...
-            .data
-            .downcast::<M>()
-            .expect("cannot downcast")
-            .into_inner();
-
-        message._touch();
-        Ok(message)
-    }
-}
+    // Called in `_read()` and `_write()` to avoid
```
GoldsteinE, line 77: I’m not sure what’s the problem `._touch()` is trying to avoid. Linked issues are related to `#[no_mangle]` / `#[used]` statics, but this is a method that’s called from other methods?

elfo-core/src/message/any.rs
```diff
@@ -0,0 +1,613 @@
...
+            marker: PhantomData,
+        }
+    }
+
+    #[inline]
+    pub fn downcast_ref<M: Message>(&self) -> Option<&'a M> {
+        let ret = self.inner.downcast_ref();
+
+        // SAFETY: we produce lifetime bound to the original one.
+        // Note: semantically it's shortening not extending.
```
GoldsteinE, line 435: I’m not sure I get why it’s shortening. It’s transmuting `'self` to `'a`, where `&'self &'a` reference exists, so `'self` can’t be shorter than `'a`.

That being said, I think this operation is sound as long as `AnyMessage` is pinned in memory, which it seems to be.

elfo-core/src/mailbox.rs
loyd: Multiple concurrent consumers are possible only in case of incorrect context usage when an actor is already done, but the context is still used from another thread. I'm not sure we should guarantee the absence of hangs here, only soundness.

elfo-core/src/envelope.rs
```diff
@@ -118,90 +200,265 @@ impl Envelope {
...
+
+        // SAFETY: `repr_ptr` is valid to write the message.
+        unsafe { ptr::replace(repr_ptr, MessageRepr::new(message)) };
+    }
+
+    fn message_repr_ptr(&self) -> NonNull<MessageRepr> {
+        let message_offset = self.header().message_offset;
+
+        // SAFETY: `message_offset` refers to the same allocation object.
+        let ptr = unsafe { self.0.as_ptr().cast::<u8>().add(message_offset as usize) };
```
GoldsteinE, line 240: `.byte_add()` can be used to avoid excessive casting.

elfo-core/src/message/repr.rs
loyd: I don't remember any test for it, only manual once, good point

elfo-core/src/message.rs
loyd: We hit this problem two years ago in the case of protocol crates without any callable code, only message declarations. The problem was that the compiler ignored `#[used]` in that case, and I'm not sure it still happens now.

Added `#[inline(never)]` callable empty `touch()` fixed this problem.

elfo-core/src/envelope.rs
```diff
@@ -118,90 +200,265 @@ impl Envelope {
...
+        (message, kind)
+    }
+
+    pub(crate) fn drop_as_unused(mut self) {
+        // SAFETY: `self` is properly initialized.
+        let header = unsafe { self.0.as_mut() };
+
+        if let MessageKind::RequestAny(token) | MessageKind::RequestAll(token) = &mut header.kind {
+            // FIXME: probably invalid for ALL requests, need to decrement remainder.
+            // REVIEW: DO NOT forget check & fix it before merging.
```
GoldsteinE, line 276: thread

elfo-core/src/message/any.rs
loyd: You're right, it's an incorrect comment

elfo-core/src/envelope.rs
loyd: oops

elfo-core/src/message.rs
GoldsteinE: So the problem is solved by calling any not-inlined function from the crate containing `#[used]` static? I don’t really understand why it’s `._read()` / `._write()` methods that do it.

elfo-core/src/exec.rs
```diff
@@ -36,6 +36,14 @@ impl ExecResult for () {
     }
 }
 
+#[sealed]
+impl ExecResult for never::Never {
+    #[allow(private_interfaces)]
+    fn unify(self) -> Result<(), BoxedError> {
+        unreachable!()
```
GoldsteinE, line 43:
```suggestion
        self
```

elfo-core/src/request_table.rs
```diff
@@ -251,7 +251,7 @@ impl ResponseToken {
 
     #[doc(hidden)]
     #[inline]
-    pub fn forget(mut self) {
+    pub fn forget(&mut self) {
```
GoldsteinE, line 254: Is there a use for `RequestToken` after forgetting it?

loyd: Actually, no. It's used only when sending to another node (semantically, a token is moved) and in `Envelope::drop_as_unused.` The second usage requires leaving `Envelope` valid (in order to run `impl Drop`), but it can be replaced with `mem::replace(header.kind, MessageKind::Regular { .. })` instead

GoldsteinE: If `.forget()` is a part of the public API (which it seems not to be, based on `#[doc(hidden)]`), I think it should take `self` and `impl Drop` should use another method

elfo-core/src/mailbox.rs
GoldsteinE: It seems like concurrent `.recv()` is in fact impossible even in theory, so this race can’t ever happen.

elfo-core/src/message/repr.rs
loyd: It's already tested in elfo-network/codec. Anyway, more tests (under miri) are added here

GoldsteinE (approved):

elfo-core/src/message/repr.rs
loyd: TLDR: It's nice idea, but leads to overcomplicated code

PR #128, opened 2024-06-05
feat(logger): implement logs truncation when line size limit exceedes

nerodono:
Adds `max_line_size` option to `logger` config. Truncation is performed in the following order:
1. `payload`
2. `meta` (timestamp, level, trace id)
3. `fields` (location, module)

### DIFF
--- a/elfo-logger/Cargo.toml
+++ b/elfo-logger/Cargo.toml
@@ -15,7 +15,6 @@ workspace = true
 
 [features]
 tracing-log = [ "dep:tracing-log", "log" ]
-docsrs = []
 
 [dependencies]
 elfo-core = { version = "0.2.0-alpha.15", path = "../elfo-core", features = ["unstable"] }
@@ -38,7 +37,7 @@ metrics = "0.17"
 dashmap = "5"
 fxhash = "0.2.1"
 humantime = "2.1.0"
-bytesize = { version = "1.2.0", features = ["serde"] }
+bytesize.workspace = true
 
 [dev-dependencies]
 elfo-core = { version = "0.2.0-alpha.15", path = "../elfo-core", features = ["test-util"] }
--- a/elfo-logger/Cargo.toml
+++ b/elfo-logger/Cargo.toml
@@ -15,6 +15,7 @@ workspace = true
 
 [features]
 tracing-log = [ "dep:tracing-log", "log" ]
+docsrs = []
 
 [dependencies]
 elfo-core = { version = "0.2.0-alpha.15", path = "../elfo-core", features = ["unstable"] }
@@ -37,6 +38,7 @@ metrics = "0.17"
 dashmap = "5"
 fxhash = "0.2.1"
 humantime = "2.1.0"
+bytesize = { version = "1.2.0", features = ["serde"] }
 
 [dev-dependencies]
 elfo-core = { version = "0.2.0-alpha.15", path = "../elfo-core", features = ["test-util"] }
--- a/elfo-logger/src/actor.rs
+++ b/elfo-logger/src/actor.rs
@@ -134,7 +134,9 @@ impl Logger {
                 || self.do_format_event::<theme::PlainTheme, TruncateOnUnfit>(&event)
         };
 
-        if !successful {
+        if successful {
+            self.shared.pool.clear(event.payload_id);
+        } else {
             unreachable!("truncation must succeed")
         }
     }
--- a/elfo-logger/src/actor.rs
+++ b/elfo-logger/src/actor.rs
@@ -19,8 +19,8 @@ use crate::{
     config::{Config, Sink},
     filtering_layer::FilteringLayer,
     formatters::Formatter,
-    line_buffer::LineBuffer,
-    line_transaction::{Line as _, LineFactory, TryDirectWrite, UseSlowPath},
+    line_buffer::{BufferConfig, LineBuffer},
+    line_transaction::{FailOnUnfit, Line as _, LineFactory, TruncateOnUnfit},
     theme, PreparedEvent, Shared,
 };
 
@@ -28,6 +28,7 @@ pub(crate) struct Logger {
     ctx: Context<Config>,
     shared: Arc<Shared>,
     filtering_layer: FilteringLayer,
+
     buffer: LineBuffer,
 }
 
@@ -54,7 +55,12 @@ impl Logger {
 
     fn new(ctx: Context<Config>, shared: Arc<Shared>, filtering_layer: FilteringLayer) -> Self {
         filtering_layer.configure(&ctx.config().targets);
-        let buffer = LineBuffer::with_buffer_capacity(1024, ctx.config().max_line_size.0 as _);
+        let buffer = LineBuffer::with_capacity(1024, {
+            let cfg = ctx.config();
+            BufferConfig {
+                max_line_size: cfg.max_line_size.0 as _,
+            }
+        });
 
         Self {
             ctx,
@@ -81,17 +87,13 @@ impl Logger {
                     let event = ward!(event, break);
                     self.buffer.clear();
 
-                    if use_colors {
-                        self.format_event::<theme::ColoredTheme, TryDirectWrite>(event)
-                    } else {
-                        self.format_event::<theme::PlainTheme, TryDirectWrite>(event)
-                    }
+                    self.format_event(use_colors, event);
 
                     if let Some(file) = file.as_mut() {
                         // TODO: what about performance here?
-                        file.write_all(self.buffer.buffer.as_bytes()).await.expect("cannot write to the config file");
+                        file.write_all(self.buffer.as_str().as_bytes()).await.expect("cannot write to the config file");
                     } else {
-                        print!("{}", self.buffer.buffer);
+                        print!("{}", self.buffer.as_str());
                     }
 
                     increment_counter!("elfo_written_events_total");
@@ -107,7 +109,10 @@ impl Logger {
                             file = open_file(self.ctx.config()).await;
                             use_colors = can_use_colors(self.ctx.config());
                             self.filtering_layer.configure(&self.ctx.config().targets);
-                            self.buffer.max_line_size = self.ctx.config().max_line_size.0 as _;
+                            self.buffer.configure(|cfg| {
+                                let actor_cfg = self.ctx.config();
+                                cfg.max_line_size = actor_cfg.max_line_size.0 as _;
+                            });
                         },
                         Terminate => {
                             // Close the channel and wait for the rest of the events.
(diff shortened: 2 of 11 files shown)


elfo-logger/src/actor.rs
```diff
@@ -133,48 +134,54 @@ impl Logger {
 
         // <timestamp> <level> [<trace_id>] <object> - <message>\t<fields>
```
loyd, line 136: In the vast majority of cases, the limit will not be reached, so I expect no performance penalty at all in such cases.

I think it better to wrap this code and call it twice:
1. With `payload=fields=buffer`
2. Check limit
3. Rerun with dedicated `payload`, `fields`, and `buffer` parts if needed

loyd (changes requested):

elfo-logger/src/line_buffer.rs
```diff
@@ -0,0 +1,147 @@
...
+
+// CurrentLine
+
+#[derive(Debug)]
+pub(crate) struct CurrentLine<'a> {
+    buffer: &'a mut LineBuffer,
+
+    // We don't wanna write unfinished data, so there must be a way to
+    // revert changes made by `CurrentLine`
+    pre_start_buffer_size: usize,
```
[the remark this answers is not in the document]
nerodono: When we, for example, wanted to implement custom line buffering algorithm, `LineBuffer` by design keeps all lines in `buffer` string, so at the moment when `CurrentLine` created, there could be data in `buffer` already

elfo-logger/src/config.rs
```diff
@@ -12,6 +14,9 @@ pub(crate) struct Config {
     #[serde(default)]
     pub(crate) format: Format,
 
+    #[serde(default)]
+    pub(crate) max_line_size: Option<MaxLineSize>,
```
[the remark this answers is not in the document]
nerodono: Yeah, the `_line_` word here used to describe limit for each line in the `buffer`, this limit applies to each line created by `CurrentLine` in the `LineBuffer`. `max_size` in that case would be wrong, since it makes this meaning even more unobvious.

elfo-logger/src/theme.rs
```diff
@@ -14,6 +14,7 @@ pub(crate) trait Theme {
     type Payload: Formatter<str>;
     type Location: Formatter<(&'static str, u32)>;
     type Module: Formatter<str>;
+    type ResetColors: Formatter<()>;
```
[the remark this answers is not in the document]
nerodono: > Where is it used?

Currently, nowhere, I'm still thinking about how to fix colored output without dirty-hacks

elfo-logger/src/line_buffer.rs
```diff
@@ -0,0 +1,147 @@
...
+}
+
+// LineBuffer
+
+#[derive(Debug)]
+pub(crate) struct LineBuffer {
+    // Buffer will contain lines and partially written meta-info, such as `<timestamp> <level>
+    // [<trace_id>]` `payload` and `fields` will be merged as soon as [`CurrentLine`] will be
+    // finished
+    pub(crate) buffer: String,
```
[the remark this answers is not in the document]
nerodono: It's not useless since `Logger` by any time can decide to clear this buffer or to write it to file/stdout, and `max_line_size` publicly needed since it can be updated when configuration changes

elfo-logger/src/actor.rs
nerodono: I suppose it's done by bd8d9cffbdc6453afa3642844ba160dd7c2755fe

elfo-logger/Cargo.toml
```diff
@@ -15,6 +15,7 @@ workspace = true
 
 [features]
 tracing-log = [ "dep:tracing-log", "log" ]
+docsrs = []
```
loyd, line 18: See elfo-dumper or elfo-core, they use `cfg(docsrs)` instead

elfo-logger/src/config.rs
```diff
@@ -12,6 +14,9 @@ pub(crate) struct Config {
     #[serde(default)]
     pub(crate) format: Format,
 
+    #[serde(default)]
```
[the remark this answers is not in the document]
loyd: Still requires comments

examples/examples/usage/config.toml
```diff
@@ -22,6 +22,9 @@
 #path = "example.log"
 #format.with_location = false
 #format.with_module = false
+## Max size of each line, in bytes, when actual line size exceedes, its contents will be
```
loyd, line 25: It should be moved to documentation of `Config`

elfo-logger/src/line_buffer.rs
```diff
@@ -0,0 +1,256 @@
...
+}
+
+impl Line for DirectWrite<'_> {
+    fn discard(self) {
+        self.0.buf.buffer.truncate(self.0.pre_start_buffer_size);
+
+        mem::forget(self);
+    }
+
+    fn total_wrote(&self) -> usize {
```
loyd, line 116: written?

elfo-logger/src/actor.rs
```diff
@@ -104,6 +107,7 @@ impl Logger {
                             file = open_file(self.ctx.config()).await;
                             use_colors = can_use_colors(self.ctx.config());
                             self.filtering_layer.configure(&self.ctx.config().targets);
+                            self.buffer.max_line_size = self.ctx.config().max_line_size.0 as _;
```
loyd, line 110: `LineBuffer` is not simply structure, it's, actually, an object because it encapsulates some behavior (not complex right now, but more invariants can be easily added in the future).
Thus, any direct access to internals should be forbidden ([LoD](https://en.wikipedia.org/wiki/Law_of_Demeter)).

* `self.buffer.configure(...)` instead of `self.buffer.max_line_size = ...`
* `self.buffer.as_bytes()` instead of `self.buffer.buffer.as_bytes()`

`self.buffer.as_bytes()` can also check (by `(debug_)assert`) implicit invariant: additional buffers (`fields` and `payload`) must be empty by that time.

elfo-logger/src/line_buffer.rs
```diff
@@ -0,0 +1,256 @@
...
+    pub(crate) fn with_buffer_capacity(capacity: usize, max_line_size: usize) -> Self {
+        Self {
+            buffer: String::with_capacity(capacity),
+            payload: String::new(),
+            fields: String::new(),
+            max_line_size,
+        }
+    }
+
+    pub(crate) const fn new(max_line_size: usize) -> Self {
```
loyd, line 187: It's actually used only in tests, so I see no reason to provide both (`with_capacity`) and `new`/`Default`.

If you still want to leave all these constructors, let's, at least, reduce initialization code by rewriting `with_capacity` in the following way:
```rust
let mut this = Self::new(max_line_size);
this.buffer.reserve(capacity);
this
```

elfo-logger/src/actor.rs
```diff
@@ -120,61 +124,79 @@ impl Logger {
...
+                    let fields_buffer = line.fields_mut();
+                    fields_buffer.push('\t');
+                    T::Module::fmt(fields_buffer, module);
+                }
             }
-        }
 
-        out.push('\n');
+            let try_slow_path =
+                !F::STOP && (line.total_wrote() > self.ctx.config().max_line_size.0 as _);
```
loyd, line 188: It looks error-prone to have logic based on `max_line_size` in two different places: here and in `LineBuffer`.

elfo-logger/src/actor.rs
```diff
@@ -120,61 +124,79 @@ impl Logger {
...
-
-        // Add ancestors' fields.
-        let mut span_id = event.span_id;
-        while let Some(data) = span_id
-            .as_ref()
-            .and_then(|span_id| self.shared.spans.get(span_id))
-        {
-            span_id.clone_from(&data.parent_id);
+        let try_slow_path = {
+            let mut line = F::create_line(&mut self.buffer);
```
loyd, line 130: I would prefer use different function for this instead of recursion as more obvious (ad, hence, less error-prone) approach:
```rust
fn format_event<T: theme::Theme>(&mut self, event: PreparedEvent) {
    let fits = self.do_format_event<T, FailOnUnfit>(&event);

    if !fits {
        self.do_format_event<T, TruncateOnUnfit>(&event);
    }

    // Cleanup
    self.shared.pool.clear(event.payload_id);
}

fn do_format_event<T: theme::Theme, L: LineLimiter>(&mut self, event: &PreparedEvent) -> bool {
    // ...
    line.try_approve()
}
```

Or even move rerun login to `LineBuffer` (that's require tracking spans of parts inside the underlying buffer), but it requires more It takes more effort.

Anyway, I can accept even the current approach if there is no desire to make code cleaner

elfo-logger/src/line_transaction.rs
```diff
@@ -0,0 +1,58 @@
...
+// This is silly fix
+pub(crate) trait LineFactory {
+    type Line<'a>: Line + 'a;
+    const STOP: bool;
+
+    fn create_line(buf: &mut LineBuffer) -> Self::Line<'_>;
+}
+
+pub(crate) struct TryDirectWrite;
+pub(crate) struct UseSlowPath;
```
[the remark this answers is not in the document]
nerodono: Done by [0719951](https://github.com/elfo-rs/elfo/pull/128/commits/07199518dc9f07d23c83b9438f2099c974519acc)

elfo-logger/src/actor.rs
nerodono: Done by [0719951](https://github.com/elfo-rs/elfo/pull/128/commits/07199518dc9f07d23c83b9438f2099c974519acc)

elfo-logger/src/line_buffer.rs
```diff
@@ -0,0 +1,274 @@
...
+        let fields_len = self.0.buf.fields.len();
+        let meta_len = self.meta_len();
+
+        let payload_part = payload_len.min(need_to_erase);
+        need_to_erase -= payload_part;
+        let fields_part = fields_len.min(need_to_erase);
+        need_to_erase -= fields_part;
+        let meta_part = meta_len.min(need_to_erase);
+
+        self.0.buf.payload.truncate(payload_len - payload_part);
```
loyd, line 87: > Panics if new_len does not lie on a [char](https://doc.rust-lang.org/stable/std/primitive.char.html) boundary.

elfo-logger/src/line_transaction.rs
```diff
@@ -0,0 +1,53 @@
...
+        buf.new_line()
+    }
+}
+
+pub(crate) trait Line {
+    fn meta_mut(&mut self) -> &mut String;
+    fn payload_mut(&mut self) -> &mut String;
+    fn fields_mut(&mut self) -> &mut String;
+
+    fn total_bytes(&self) -> usize;
```
loyd, line 51: Is it used outside `line_buffer.rs`?

elfo-logger/src/line_buffer.rs
```diff
@@ -0,0 +1,274 @@
...
+        self.0.buf.fields.truncate(fields_len - fields_part);
+        self.0
+            .buf
+            .buffer
+            .truncate(self.0.buf.buffer.len() - meta_part);
```
loyd, line 92: What about adding some info that the line is truncated? For instance, in the dumper, we add " TRUNCATED"

elfo-logger/src/line_buffer.rs
```diff
@@ -0,0 +1,274 @@
+use std::mem;
+
+use crate::line_transaction::Line;
+
+// BufferConfiguration
```
loyd, line 5: `BufferConfig`?

Also, everywhere in elfo sections are formatted as `// === Name ===`, but it's not required, so you can resolve this comment

elfo-logger/src/line_buffer.rs
```diff
@@ -0,0 +1,274 @@
...
+
+    // We don't wanna write unfinished data, so there must be a way to
+    // revert changes made by `CurrentLine`
+    pre_start_buffer_size: usize,
+}
+
+// CurrentLine
+
+#[derive(Debug)]
+pub(crate) struct CurrentLine<'a>(Repr<'a>);
```
loyd, line 34: Rename to something more meaningful?

elfo-logger/src/actor.rs
```diff
@@ -120,61 +129,81 @@ impl Logger {
...
-        T::ActorMeta::fmt(out, &event.object);
-        out.push_str(" - ");
-        T::Payload::fmt(out, &payload);
+        T::Timestamp::fmt(line.meta_mut(), &event.timestamp);
+        line.meta_mut().push(' ');
+        T::Level::fmt(line.meta_mut(), event.metadata.level());
+        line.meta_mut().push_str(" [");
+        T::TraceId::fmt(line.meta_mut(), &event.trace_id);
+        line.meta_mut().push_str("] ");
+        T::ActorMeta::fmt(line.payload_mut(), &event.object);
```
loyd, line 165: Why not `meta_mut` here?

elfo-logger/src/line_buffer.rs
```diff
@@ -0,0 +1,274 @@
...
+}
+
+// DirectWrite
+
+#[derive(Debug)]
+pub(crate) struct DirectWrite<'a>(Repr<'a>);
+
+impl<'a> Drop for DirectWrite<'a> {
+    fn drop(&mut self) {
+        self.0.buf.buffer.push('\n');
```
loyd, line 116: I think it's supposed to be in `try_commit`

elfo-logger/src/actor.rs
nerodono: Since `event.object` and `payload` are part of the payload, originally

elfo-logger/src/line_buffer.rs
nerodono: Done in [c2495f0](https://github.com/elfo-rs/elfo/pull/128/commits/c2495f0651fe992875ff299919774d87a0905418)

elfo-logger/src/line_buffer.rs
nerodono: Done in [c2495f0](https://github.com/elfo-rs/elfo/pull/128/commits/c2495f0651fe992875ff299919774d87a0905418)

elfo-logger/src/line_transaction.rs
nerodono: `total_bytes` nowhere, actually, removed it from the trait in [c2495f0](https://github.com/elfo-rs/elfo/pull/128/commits/c2495f0651fe992875ff299919774d87a0905418)

elfo-logger/src/line_buffer.rs
nerodono: Implemented `safe_truncate` function in [0d59222](https://github.com/elfo-rs/elfo/pull/128/commits/0d592224438b8e92bcf884ad50921678f3346d22)

loyd (approved):

PR #138, opened 2024-11-01
chore: derive ActorStatusReport constructor

Erigara:
Add constructor to `ActorStatusReport` under `test-util` feature flag.

### DIFF
--- a/elfo-core/src/messages.rs
+++ b/elfo-core/src/messages.rs
@@ -113,3 +113,13 @@ pub struct ActorStatusReport {
     pub meta: Arc<ActorMeta>,
     pub status: ActorStatus,
 }
+
+impl ActorStatusReport {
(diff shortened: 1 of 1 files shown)


elfo-core/src/messages.rs
```diff
@@ -108,6 +108,7 @@ impl SubscribeToActorStatuses {
 }
 
 #[message]
+#[cfg_attr(feature = "test-util", derive(Constructor))]
```
nerodono, line 111: Looks like deriving constructor makes `#[non_exhaustive]` useless for creation of the type.
Writing specific constructor by hand seems more meaningful, since it won't (or shouldn't) update as definition of `ActorStatusReport` grows

Erigara: Added by hand constructor.

PR #140, opened 2024-11-26
Add ability to change network discovery configuration on the fly

nerodono:
This PR adds ability to change some `elfo-network` configuration aspects on the fly, more precisely - discovery settings.

### DIFF
--- a/elfo-network/src/discovery/mod.rs
+++ b/elfo-network/src/discovery/mod.rs
@@ -178,7 +178,7 @@ impl Discovery {
         let capabilities = self.get_capabilities();
 
         for transport in &self.cfg.listen {
-            let stream = socket::listen(&transport, node_no, launch_id, capabilities)
+            let stream = socket::listen(transport, node_no, launch_id, capabilities)
                 .await
                 .wrap_err_with(|| eyre!("cannot listen {}", transport))?
                 .filter_map(move |socket| async move {
@@ -669,10 +669,10 @@ mod tests {
                 let (diff_new, diff_removed) = (expected.0.as_ref(), expected.1.as_ref());
 
                 Self {
-                    old: old.iter().copied().collect(),
-                    new: new.iter().copied().collect(),
+                    old: old.to_vec(),
+                    new: new.to_vec(),
                     expected: Diff {
-                        new: diff_new.iter().copied().collect(),
+                        new: diff_new.to_vec(),
                         removed: diff_removed.iter().copied().collect(),
                     },
                 }
--- a/elfo-network/src/discovery/mod.rs
+++ b/elfo-network/src/discovery/mod.rs
@@ -19,6 +19,8 @@ use crate::{
     NetworkContext,
 };
 
+use diff::Diff;
+
 /// Initial window size of every flow.
 /// TODO: should be different for groups and actors.
 const INITIAL_WINDOW_SIZE: i32 = 100_000;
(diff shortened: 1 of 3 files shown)


elfo-network/src/discovery/mod.rs
```diff
@@ -550,6 +584,37 @@ async fn recv_regular<M: Message>(socket: &mut Socket, idle_timeout: Duration) -
     })
 }
 
+#[cfg_attr(test, derive(Debug))]
+struct Diff<T> {
```
loyd, line 588: Let's move it to a dedicated module (`discovery/diff.rs`)

elfo-network/src/discovery/mod.rs
```diff
@@ -550,6 +584,37 @@ async fn recv_regular<M: Message>(socket: &mut Socket, idle_timeout: Duration) -
...
+    new: Vec<T>,
+    removed: fxhash::FxHashSet<T>,
+}
+
+fn diff_owned<'a, T, I>(old: fxhash::FxHashSet<T>, new: I) -> Diff<T>
```
loyd, line 593: What about `Diff::new(old: impl IntoIterator, new: impl IntoIterator)`?

nerodono: IMO better name would be `Diff::make`, and, about making `old` IntoIterator instead: I think showing `HashSet` in constructor is better, since it highlights that calling this API may be not that efficient (first you need to build hash set, then operate on it), gives caller opportunity to optimize diffs (by constructing hashset from ground up, for example). While these points are about optimization and not so important here, however, leaving opportunities here is easier than not leaving, so I decided to design API this way.

elfo-network/src/discovery/mod.rs
nerodono: Done

elfo-network/src/discovery/mod.rs
```diff
@@ -561,3 +597,5 @@ fn unexpected_message_error(envelope: Envelope, expected: &[&str]) -> eyre::Repo
 async fn timeout<T>(duration: Duration, fut: impl Future<Output = Result<T>>) -> Result<T> {
     tokio::time::timeout(duration, fut).await?
 }
+
+mod diff;
```
loyd, line 601: move after `use` statements (and, optionally, replace `diff::X` with more explicit `self::diff::X`)

elfo-network/src/discovery/mod.rs
```diff
@@ -132,19 +135,52 @@ impl Discovery {
...
+                predefined.into_iter().collect(),
+                self.ctx.config().discovery.predefined.iter(),
+            );
+            for transport in new {
+                self.discover(transport);
+            }
+
+            // FIXME: handle removal.
+            if !removed.is_empty() {
+                error!(
```
loyd, line 169: I think this should be `warn!`

loyd (approved):

elfo-network/src/discovery/mod.rs
```diff
@@ -132,19 +135,52 @@ impl Discovery {
...
+        self.update_discovery(old.discovery);
+    }
+
+    fn update_discovery(&mut self, old: config::DiscoveryConfig) {
+        let config::DiscoveryConfig {
+            predefined,
+            attempt_interval: _,
+        } = old;
+
+        {
```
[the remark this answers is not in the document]
nerodono: To visually distinguish steps (i.e. updating interval, when changing this property on the fly would be implemented). Primary reason was due to possible shadowing in the next revisions, it's easier to name things meaningfully and control their liveness in separate scopes.

elfo-network/src/discovery/mod.rs
nerodono: Done

PR #142, opened 2024-12-08
feat(core): implement `Context::send` in the builder-like manner

nerodono:
Imperfect draft of turning `Context::send` into chained builder. Adding behavior settings for `Context::send` could be annoying, we already have functions like `unbounded_send`, `send`, `send_to`, `unbounded_send_to`, `try_send`, `try_send_to` and so on. With turning `send` into builder, we could achieve having only one function in the `Context`: `Context::send`, actions performed on that builder could be divided in two categories

## Outcome

Function which finishes building of envelope and sends it, such as
- `ctx.send(..).try_()` - try send without waiting for available place in the mailbox
- `ctx.send(..).unbounded()` - send envelope unconditionally
- `ctx.send(..).await` - wait for place in the destination mailbox and send

## Parameters

- `.to(..)` - destination

Moreover, we could elide `request` and `request_to` functions as well, but it requires more thought though, since `ctx.send(..).resolve().await` looks nice, but `ctx.send(..).all().await` seems confusing (Maybe unifying `send` and `request` interfaces is quite unfortunate?)

### DIFF
--- a/elfo-core/src/context.rs
+++ b/elfo-core/src/context.rs
@@ -33,6 +33,7 @@ use crate::{
 use self::stats::Stats;
 
 mod stats;
+mod send;
 
 static DUMPER: Lazy<Dumper> = Lazy::new(|| Dumper::new(INTERNAL_CLASS));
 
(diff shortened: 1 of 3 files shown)


elfo-core/src/context.rs
```diff
@@ -61,6 +63,238 @@ enum Stage {
...
+    ///
+    /// // Fire or log.
+    /// if let Err(error) = ctx.try_send(SomethingHappened) {
+    ///     tracing::warn!(%error, "...");
+    /// }
+    /// # }
+    /// ```
+    ///
+    /// [inter-group routing]: https://actoromicon.rs/ch04-01-routing.html
+    pub fn try_send<M: Message>(&self, message: M) -> Result<(), TrySendError<M>> {
```
nerodono, line 186: Maybe deprecate those methods in favor of just send?

elfo-core/src/context/builder.rs
```diff
@@ -0,0 +1,206 @@
+use std::future::{Future, IntoFuture};
+
+use idr_ebr::EbrGuard;
+use tracing::trace;
+
+use crate::{
+    context::{addrs_with_envelope, e2m, DUMPER},
```
nerodono, line 7: I guess some of this functionality now should be here, but I'm quite reluctant to move more code more at this state.

PR #145, opened 2024-12-16
feat(dumper): implement `time` in dump path template

nerodono:
Implements more flexible templating for storing dumps. Questions needed to be resolved in order to undraft:

- [x] Time precision

The `{time}` itself has no information about precision (to hours? to minutes? to seconds?) and format (ISO? unix time?), we either need to choose format and precision, or allow customization, like  `strftime`, but simpler: `{time:dd-mm-YYYY}` with some default when format is omitted.
IMO latter is better, it gives more flexibility in granularity.

- [x] Time source

We could use last dump message time or time when writing dump to the disk. IMO latter is nice approximation, since for the rotation purposes we usually need to clean-up space, not specific time interval and using that approximation wouldn't be problematic - time of writing the dump is never less than last dump message time, so, the worst-case scenario is keeping for some time expired files in FS. Or the hell scenario is broken system clock, which is problematic in any chosen way.

### DIFF
--- a/elfo-dumper/src/actor.rs
+++ b/elfo-dumper/src/actor.rs
@@ -1,4 +1,8 @@
-use std::{iter, panic, sync::Arc, time::Duration};
+use std::{
+    iter, panic,
+    sync::Arc,
+    time::{Duration, SystemTime},
+};
 
 use eyre::{Result, WrapErr};
 use fxhash::FxHashSet;
@@ -20,7 +24,7 @@ use elfo_core::{
 use elfo_utils::ward;
 
 use crate::{
-    config::Config,
+    config::{dump_path::TemplateVariables, Config},
     dump_storage::{Drain, DumpRegistry, DumpStorage},
     file_registry::{FileHandle, FileRegistry},
     reporter::{Report, Reporter},
@@ -80,8 +84,32 @@ impl Dumper {
         }
     }
 
+    fn make_template_variables(&self) -> TemplateVariables<'_> {
+        let now = SystemTime::now();
+        let ts = now
+            .duration_since(SystemTime::UNIX_EPOCH)
+            .expect("shit happens")
+            .as_secs();
+
+        TemplateVariables {
+            class: self.ctx.key(),
+            ts,
+        }
+    }
+
+    fn render_path(&self, to: &mut String) {
+        to.clear();
+        self.ctx
+            .config()
+            .path
+            .render_into(self.make_template_variables(), to);
+    }
+
     async fn main(mut self) -> Result<()> {
-        let mut path = self.ctx.config().path(self.ctx.key());
+        let mut path = String::new();
+        let mut path_swap = String::new();
+
+        self.render_path(&mut path);
         self.file_registry
             .open(&path, false)
             .await
@@ -106,7 +134,7 @@ impl Dumper {
                     let config = self.ctx.config();
                     self.interval.set_period(config.write_interval);
 
-                    path = config.path(self.ctx.key());
+                    self.render_path(&mut path);
                     self.file_registry
                         .open(&path, false)
                         .await
@@ -132,7 +160,16 @@ impl Dumper {
                 DumpingTick => {
                     let timeout = self.ctx.config().write_interval;
                     let dump_registry = self.dump_registry.clone();
-                    let file = self.file_registry.acquire(&path).await;
+
+                    // NOTE: could be optimized by not re-rendering path
+                    // if variables aren't changed in the affectable way, it's
+                    // not that matters here though.
+                    self.render_path(&mut path_swap);
+                    let file = self.file_registry.acquire_for_write(&path, &mut path_swap).await?;
+                    if path != path_swap {
+                        path.clear();
+                        std::mem::swap(&mut path, &mut path_swap);
+                    }
 
                     // A blocking background task that writes a lot of dumps in batch.
                     // It's much faster than calling tokio's async functions.
--- a/elfo-dumper/src/actor.rs
+++ b/elfo-dumper/src/actor.rs
@@ -89,7 +89,7 @@ impl Dumper {
         let ts = now
             .duration_since(SystemTime::UNIX_EPOCH)
             .expect("shit happens")
-            .as_secs();
+            .as_secs() as i64;
 
         TemplateVariables {
             class: self.ctx.key(),
@@ -165,7 +165,10 @@ impl Dumper {
                     // if variables aren't changed in the affectable way, it's
                     // not that matters here though.
                     self.render_path(&mut path_swap);
-                    let file = self.file_registry.acquire_for_write(&path, &mut path_swap).await?;
+                    let file = self
+                        .file_registry
+                        .acquire_for_write(&path, &mut path_swap)
+                        .await?;
                     if path != path_swap {
                         path.clear();
                         std::mem::swap(&mut path, &mut path_swap);
--- a/elfo-dumper/src/actor.rs
+++ b/elfo-dumper/src/actor.rs
@@ -167,7 +167,7 @@ impl Dumper {
                     self.render_path(&mut path_swap);
                     let file = self
                         .file_registry
-                        .acquire_for_write(&path, &mut path_swap)
+                        .acquire_for_write(&path, &path_swap)
                         .await?;
                     if path != path_swap {
                         path.clear();
--- a/elfo-dumper/src/config.rs
+++ b/elfo-dumper/src/config.rs
@@ -10,6 +10,10 @@ use std::time::Duration;
 use bytesize::ByteSize;
 use serde::Deserialize;
 
+pub(crate) mod dump_path;
+
+pub use dump_path::DumpPath;
+
 /// The dumper's config.
 ///
 /// # Examples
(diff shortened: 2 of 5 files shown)


elfo-dumper/src/config/dump_path.rs
```diff
@@ -0,0 +1,259 @@
...
+
+            Self::push_path(&mut plain_size, &mut components);
+            components.push(Component {
+                // "{<variable>}"
+                size: raw_variable.len() as u16 + 2,
+                data: ComponentData::Variable(variable),
+            });
+            chunk = after;
+        }
+        if !chunk.is_empty() {
```
nerodono, line 136: This is not needed, since the loop can only exit through `chunk = ""`

Cargo.toml
```diff
@@ -33,6 +33,7 @@ metrics = "0.17.1"
 dashmap = "6.0.1"
 toml = "0.8.14"
 bytesize = { version = "1.2.0", features = ["serde"] }
+test-case = "3.3.1"
```
loyd, line 36: dev

elfo-dumper/src/actor.rs
```diff
@@ -106,7 +109,10 @@ impl Dumper {
                     let config = self.ctx.config();
                     self.interval.set_period(config.write_interval);
 
-                    path = config.path(self.ctx.key());
+                    path = config.path.render(TemplateVariables {
```
loyd, line 112: We should rerender the path on every `DumpingTick`

elfo-dumper/src/config/dump_path.rs
```diff
@@ -0,0 +1,259 @@
...
+        vars: TemplateVariables<'_>,
+        expected: &str,
+    ) {
+        let path = DumpPath::parse(template);
+        let actual = path.render(vars);
+
+        assert_eq!(actual, expected);
+    }
+
+    #[test_case("/tmp/dump.dump", [
```
loyd, line 217: `test_case` seems to be not very helpful here:
* extra dependency
* generates a lot of tiny tests
* the number of lines will be the same even without `test_case`

elfo-dumper/src/config/dump_path.rs
nerodono: Interestingly, if I remove that branch, rust itself says that assignment in `chunk = ""` is not used (no reads after this operation). Kinda surprised how compiler sometimes not that dumb.

elfo-dumper/src/actor.rs
nerodono: Done, there's some cost in that though, but it's not like that matters here

elfo-dumper/src/config/dump_path.rs
nerodono: Implemented simple DSL without the `test_case`, looks nicer and rustfmtable

nerodono:
Holding file opened looks strange if we have dynamic parts in the template, since open time and time when happens `DumpTick` are different, so holding open e.g. `system.network-1337.dump` is not that meaningful, since we'll open new file any way. There's two possible solutions for this:
1. Divide template variables in two categories: dynamic and static. Static template variables could be pre-rendered, dynamic parts are kept as-is, this pre-rendered template with dynamic parts could be stored as a key in file registry.  Can be summarized as - allow reserving whole "namespaces" of files instead of specific files.
2. Open files every time on `DumpTick` - simplest way

Doing things smart way (1) is even more harder if we take precision into account, for example, for file `/etc/{time:YYYY}` IS meaningful to hold file open, since year is not a thing that changes really fast. There's third option, hybrid of these two - if template contains any dynamic parts, open file every time on `DumpTick`, otherwise, hold files open, however, this options suffers from the first problem too (`/etc/{time:YYYY}` changes rarely), but can be worked-around by making dynamicity check smarter, like holding file open if maximum time precision is less than a hour (less = less precise, greater time steps) and relate it somehow to dump tick interval.

elfo-dumper/src/config/dump_path.rs
```diff
@@ -0,0 +1,273 @@
...
+        );
+
+        // Class and time.
+        case(
+            "/tmp/dump-{class}-{time}.dump",
+            TemplateVariables {
+                class: "class",
+                ts: 1337100,
+            },
+            "/tmp/dump-system.network-1337100.dump",
```
nerodono, line 214: class

elfo-dumper/src/config.rs
```diff
@@ -39,7 +43,7 @@ pub struct Config {
     /// A path to a dump file or template:
     /// * `path/all.dump` - one file.
     /// * `path/{class}.dump` - file per class.
-    pub path: String,
+    pub path: DumpPath,
```
loyd, line 46: documentation

nerodono:
> Holding file opened looks strange if we have dynamic parts in the template, since open time and time when happens `DumpTick` are different, so holding open e.g. `system.network-1337.dump` is not that meaningful, since we'll open new file any way. There's two possible solutions for this:
> 
>     1. Divide template variables in two categories: dynamic and static. Static template variables could be pre-rendered, dynamic parts are kept as-is, this pre-rendered template with dynamic parts could be stored as a key in file registry.  Can be summarized as - allow reserving whole "namespaces" of files instead of specific files.
> 
>     2. Open files every time on `DumpTick` - simplest way
> 
> 
> Doing things smart way (1) is even more harder if we take precision into account, for example, for file `/etc/{time:YYYY}` IS meaningful to hold file open, since year is not a thing that changes really fast. There's third option, hybrid of these two - if template contains any dynamic parts, open file every time on `DumpTick`, otherwise, hold files open, however, this options suffers from the first problem too (`/etc/{time:YYYY}` changes rarely), but can be worked-around by making dynamicity check smarter, like holding file open if maximum time precision is less than a hour (less = less precise, greater time steps) and relate it somehow to dump tick interval.

I guess we're okay as for now with whatever meaning of time used, for now it's time of opening the file or time of writing to the disk, depends on what precision of time is used, for most use-cases, where user doesn't use seconds precision, it would be time of opening the file.

nerodono:
There's some unsafe dropped in the place, it's used only because of using `libc` for formatting the time. There's no small libraries implementing that feature, or there's no library that allows opt-outing most of the features so that build complexity isn't grown significantly. libc is definitely on user's build tree, so it makes no difference.

Unsafe is used to:
1. Use functions `localtime_r`, `strftime` itself
2. `CString` interactions

unsafe usages are isolated and tested, so I guess it's okay

elfo-dumper/src/config/dump_path.rs
```diff
@@ -0,0 +1,427 @@
...
+#[derive(Debug, Clone)]
+#[cfg_attr(test, derive(PartialEq, Eq))]
+enum Variable {
+    /// Dump class.
+    Class,
+
+    /// Time of the dump.
+    Time {
+        /// strptime string format.
+        format: cstr::Utf8CString,
```
nerodono, line 194: strftime

elfo-dumper/src/config/dump_path.rs
```diff
@@ -0,0 +1,427 @@
...
+
+#[derive(Debug, Clone)]
+#[cfg_attr(test, derive(PartialEq, Eq))]
+enum Variable {
+    /// Dump class.
+    Class,
+
+    /// Time of the dump.
+    Time {
+        /// strptime string format.
```
nerodono, line 193: strftime

elfo-dumper/src/config/dump_path.rs
```diff
@@ -0,0 +1,427 @@
...
+    };
+    if len == 0 {
+        return false;
+    }
+
+    // SAFETY: comments inside.
+    unsafe {
+        // 1. mem::transmute wouldn't compile if u8 and libc::c_char differ in size
+        // (there are strange circumstances).
+        // 2. `libc::c_char` to i8 is i8 -> u8 conversion, thus converting
```
nerodono, line 244: `libc::c_char` to u8

elfo-dumper/src/actor.rs
```diff
@@ -80,8 +84,32 @@ impl Dumper {
         }
     }
 
+    fn make_template_variables(&self) -> TemplateVariables<'_> {
+        let now = SystemTime::now();
+        let ts = now
+            .duration_since(SystemTime::UNIX_EPOCH)
+            .expect("shit happens")
```
loyd, line 91: `.unwrap_or(SystemTime::UNIX_EPOCH)` or more appropriate comment

elfo-dumper/src/actor.rs
```diff
@@ -106,7 +134,7 @@ impl Dumper {
                     let config = self.ctx.config();
                     self.interval.set_period(config.write_interval);
 
-                    path = config.path(self.ctx.key());
+                    self.render_path(&mut path);
                     self.file_registry
```
loyd, line 138: If the active file is changed between reconfiguration and the following `DumpingTick`, we leave the previous one unclosed.

It seems we should simply remove `fn open()` here and always open files lazily

elfo-dumper/src/file_registry.rs
```diff
@@ -17,30 +17,56 @@ pub(crate) struct FileRegistry {
...
+        Self::open_inner(path, force, &mut file).await?;
+        Ok(())
+    }
+
+    /// Open file for writing dump in it, closing previous one if
+    /// paths differ.
+    pub(crate) async fn acquire_for_write(&self, old_path: &str, path: &str) -> Result<FileHandle> {
+        let mut fh = {
+            let mut files = self.files.lock();
+            if old_path != path {
```
loyd, line 52: We can move the open logic to `FileHandle` and store `FileHandle` for a long time in every actor (instead of `path` and `path_swap`), which makes the code both clearer and less error-prone

PR #146, opened 2024-12-17
Remove lifetime from `Proxy::request()` future

Arjentix:
- Remove lifetime from `Proxy::request()`, which helps in testing complex request scenarios
- Fix CI


sargarass:
`request_to` should also be covered by this change

loyd (approved):

PR #147, opened 2024-12-19
fix: sanitize Debug impls against other possible traits with `fmt`

BulbaWarrior:
If a type implemets some other trait containing `fn fmt(..)`, current `Debug` implementation becomes ambiguous.
This disambiguates `fmt` calls


loyd (approved):
Thanks for your contribution!

PR #148, opened 2024-12-23
allow construction of arbitrary `ActorStatusKind` in tests

BulbaWarrior:
Not all `ActorStatusKind` variants are public, however it is sometimes desirable to construct arbitrary variants in testing environment

### DIFF
--- a/elfo-core/src/actor_status.rs
+++ b/elfo-core/src/actor_status.rs
@@ -17,22 +17,20 @@ pub struct ActorStatus {
 
 impl ActorStatus {
     pub const ALARMING: ActorStatus = ActorStatus::new(ActorStatusKind::Alarming);
-    pub(crate) const FAILED: ActorStatus = ActorStatus::new(ActorStatusKind::Failed);
-    pub const INITIALIZING: ActorStatus = ActorStatus::new(ActorStatusKind::Initializing);
-    pub const NORMAL: ActorStatus = ActorStatus::new(ActorStatusKind::Normal);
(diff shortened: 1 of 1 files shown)


elfo-core/src/actor_status.rs
```diff
@@ -23,13 +23,22 @@ impl ActorStatus {
...
+    #[cfg(not(feature = "test-util"))]
     const fn new(kind: ActorStatusKind) -> Self {
         Self {
             kind,
             details: None,
         }
     }
 
+    #[cfg(feature = "test-util")]
+    pub const fn new(kind: ActorStatusKind) -> Self {
```
loyd, line 35: What is the reason for this change? Is it to create `FAILED` and `TERMINATED` statuses in tests?

BulbaWarrior: yes

loyd: I think it's more consistent to extend the visibility of `ActorStatus::X` costs under `test-util` instead of providing another public way to create `ActorStatus.`

BulbaWarrior: fixed

elfo-core/src/actor_status.rs
```diff
@@ -17,10 +17,17 @@ pub struct ActorStatus {
 
 impl ActorStatus {
     pub const ALARMING: ActorStatus = ActorStatus::new(ActorStatusKind::Alarming);
+
+    #[cfg(not(feature = "test-util"))]
```
BulbaWarrior, line 21: maybe it will look better if we add two separate impl blocks for consts under different features?

PR #149, opened 2025-01-14
fix(network/worker): forget response token to prevent recursion on drop

Erigara:
It was observed that in case of panic on other side of network actor, response token ended up in endless recursion due `Drop` implementation.

### DIFF
--- a/elfo-network/src/worker/mod.rs
+++ b/elfo-network/src/worker/mod.rs
@@ -891,17 +891,24 @@ impl remote::RemoteHandle for RemoteHandle {
 
         let recipient = NetworkAddr::from_remote(token.sender());
 
+        let mut item = Some(KanalItem {
+            recipient,
+            envelope,
+            token: Some(token),
+        });
+
         if likely(self.tx_flows.do_acquire(recipient)) {
-            let item = KanalItem {
-                recipient,
-                envelope,
-                token: Some(token),
-            };
-            match self.tx.try_send(item) {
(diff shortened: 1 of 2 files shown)


sargarass:
I'm not sure this PR fixes the issue. We probably don't want to re-enter the drop with the same ResponseToken twice. This means we either check all possible outcomes and manually handle them with forget (which is error-prone and potentially dangerous in [this](https://github.com/elfo-rs/elfo/pull/149/commits/98537b5bee23833afca2dc049c4dd957bf3cb579#diff-db1fb2440946424e713b230269ce45683f066a1312f41561dd748a6a17534580R903) case too and down the line), or we could use some kind of bitflags to mark tokens as responded.

Erigara:
Ok, i agree, setting some kind of flag would be more robust approach in this case.

elfo-network/src/worker/mod.rs
```diff
@@ -902,6 +902,9 @@ impl remote::RemoteHandle for RemoteHandle {
                 Ok(false) => unreachable!(),
                 Err(_) => {}
             }
+        } else {
+            // Forget token to prevent recursion on drop
+            token.forget();
```
loyd, line 907: I'm sure this should be called unconditionally before `trace!`. And.. the better comment is required

Erigara: Fixed + i've added test to reproduce the issue.

PR #150, opened 2025-01-16
feat(network): divide compression algorithm support in three categories

nerodono:
This PR introduces more flexibility in compression algorithm selection by dividing "support" in three categories:
- `supported` - the algorithm can be used to communicate with the node
- `preferred` - the algorithm should be selected to communicate with the node
- `disabled` - the algorithm is not supported

Two nodes decide on the algorithm as follows:
- If they both have preferred algorithms in common - select in [hard-coded] preference order
- If first node prefers something that is supported by the second node - the algorithm is selected in the [hard-coded] preference order
- Otherwise, no compression would be used

First of all, this algorithm implies there's node which "leads" in choice of algorithm ("first node"), in the current implementation, lead node is node that accepts the connection.

**BREAKING CHANGES**: compression algorithms in config expressed now as follows:
```toml
algorithms.lz4 = { preference = "Preferred" }
algorithms.zstd = { preference = "Disabled" }
```


elfo-network/src/socket/mod.rs
```diff
@@ -301,10 +489,21 @@ mod tests {
...
+    const EMPTY_CAPS: Capabilities = Capabilities::new(Compression::empty());
+
     #[message]
     #[derive(PartialEq)]
     struct TestSocketMessage(String);
 
+    #[test]
+    fn capabilities_format_is_compatible_with_020alpha17() {
+        let caps = Capabilities::new(Compression::new(Algorithms::LZ4, Algorithms::empty()));
+        dbg!(caps.bits());
```
nerodono, line 501: dbg

elfo-network/src/socket/mod.rs
```diff
@@ -25,9 +25,194 @@ mod idleness;
 mod raw;
 
 bitflags::bitflags! {
-    #[derive(Clone, Copy)]
-    pub(crate) struct Capabilities: u32 {
-        const LZ4 = 1 << 8;
+    #[derive(Debug, Clone, Copy, PartialEq, Eq)]
+    pub(crate) struct Algorithms: u8 {
+        const LZ4 = 1;
+        // NB: Shift by 2: `const ZSTD = 1 << 2;`.
```
nerodono, line 31: New algorithm should be `1` and lz4 is `1 << 2`, since shifting to 9th bit would interpret previously lz4 bit as a `zstd` bit. TBH adding new algorithm is quite tiring, maybe we could reserve all (8) bits for compression algorithms, this removes need for the `NO_ELEMS` constant.
This would make a hole in lowest bit since elfo initially for some reason has chosen 8th bit instead of 7th (counting from zero) for lz4, lowest bit still can be used though.

loyd: > Reserving whole 8 bits would make a hole in lowest bit

Why?

> since elfo initially for some reason has chosen 8th bit instead of 7th

I don't get this. It's literally `Capabilities { compression: uXX, rest: u8 }` to allow further convenient composition. There were no reasons to use `<< 7` here.

loyd: I see no point in having such `Algorithms`. The next codec must be added with 2-bit shifting, so this is a slightly broken abstraction. It makes more sense (and a more obvious approach) to define `LZ4_SUPPORTED = 1 << 1; LZ4_PREFERRED = 1` directly (or, even simpler, to remove this `Algorithms` and encode everything in `Capabilities`).

Honestly, I don't like the current overcomplicated code: `NO_ELEMS`, `COMPRESSION_SHIFT`, and so on.
I prefer a direct simple approach because we have only one capability right now.

Further, by adding other planned capabilities (additional checksums, protocol-name compaction, TLS, and so on), it should be replaced. I like the approach used in sharded-slab https://github.com/hawkw/sharded-slab/blob/e540cdb7daafd6a6e9d17408d3de017029a6637f/src/lib.rs#L1026 for packing, but right now this is overkill.

loyd (changes requested):

elfo-network/src/config.rs
```diff
@@ -58,22 +58,46 @@ pub struct Config {
...
+/// Preference.
+#[derive(Debug, Clone, Copy, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum Preference {
+    /// This is preferred.
+    Preferred,
+
+    /// This is just supported.
+    Supported,
+}
```
loyd, line 70: Probably we should also add a direct `Disabled` option here

elfo-network/src/socket/handshake.rs
```diff
@@ -39,6 +41,18 @@ impl Handshake {
         }
     }
 
+    pub(super) fn choose_compression(&self) -> Option<CompressionAlgorithm> {
```
loyd, line 44: I'm sure all helpers should be defined on `capabilities`. Otherwise, any new capabilities will also require an additional method here, but it's pointless while `Capabilities` is reachable directly (because this is a public field of `Handshake`).

elfo-network/src/config.rs
```diff
@@ -58,22 +58,46 @@ pub struct Config {
...
+    ///
+    /// Preferred implies supported.
     #[serde(default)]
-    pub algorithm: CompressionAlgorithm,
+    pub algorithms: CompressionPreference,
 }
 
 /// Compression algorithms.
 #[derive(Debug, Default, PartialEq, Eq, Deserialize, Clone)]
 pub enum CompressionAlgorithm {
```
loyd, line 97: This is not a part of the config anymore, there is no reason to have it in this module and implement `Deserialize` for this.

elfo-network/src/socket/mod.rs
```diff
@@ -25,9 +25,194 @@ mod idleness;
...
+    pub(crate) const fn preferred(self) -> Algorithms {
+        // `supported` bits would be discarded.
+        Algorithms::from_bits_truncate(self.0)
+    }
+}
+
+/// Layout:
+/// ```text
+///        Bits
+///    23    2     7
```
loyd, line 168: I'm confused here. Why seven lowest bits are reserved, not eight?
* 22 bits reserved
* 1 bit for "LZ4_SUPPORTED"
* 1 bit for "LZ4_PREFERRED"
* 8 bits reserved

elfo-network/src/config.rs
```diff
@@ -58,22 +58,46 @@ pub struct Config {
...
 
+/// Preference.
+#[derive(Debug, Clone, Copy, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum Preference {
+    /// This is preferred.
+    Preferred,
+
+    /// This is just supported.
+    Supported,
```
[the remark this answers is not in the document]
loyd: However, the default can vary for different capabilities in the future (e.g. protocol-name compaction will be "Preferred" by default). So, it's better to specify a default for `lz4`, not the whole `Preference` as said in the previous comment.

elfo-network/src/config.rs
```diff
@@ -58,22 +58,46 @@ pub struct Config {
...
+    ///
+    /// Example:
+    /// ```toml
+    /// algorithms = { lz4 = "preferred" }
+    /// ```
+    ///
+    /// Preferred implies supported.
     #[serde(default)]
-    pub algorithm: CompressionAlgorithm,
+    pub algorithms: CompressionPreference,
```
loyd, line 92: This is a breaking change (which is ok because we have alpha versions), but, in this case, I think we can revise it even more and use `system.network.compression.lz4 = "Preferred"`. If other settings (e.g. compression level) will be available to configure in the future, it will support the second per-codec form `system.network.compression.zstd = { preference = "Preferred", level = 6 }` in a backward-compatible way. I don't see what common options can be added next to `algorithms` to have this additional nesting level.

elfo-network/src/socket/mod.rs
nerodono: > Why?

I guess I got wrong idea from the code, allocating 24bits for compression algorithm is kinda unexpected for me.

> this is a slightly broken abstraction.

It's not, since `Algorithms` is not even an abstraction, it's just a fancy wrapper around `u8` (I was writing code with that in mind, `Algorithms` aren't designed to encapsulate or hide something). But also, if we want to call that an abstraction, the real abstraction here is not the `Algorithms`, but a `Compression` - only `Compression` knows and uses the underlying type layout information and hides behind the scenes how it arranges these bits.

> `NO_ELEMS` and `COMPRESSION_SHIFT`

First of all, the need in these two were dictated by my misunderstanding of intention of previously written code. Secondly, `NO_ELEMS` is kinda harsh hack to not contribute to `bitflags` at the moment (see the comment nearby), so I expect this code to wander around as long as my PR (which I have to make) to the `bitflags` wouldn't be merged and published to the crates.io.

So yes, these things aren't needed, since they here because I misunderstood intention of previously written code, in the first place.

elfo-network/src/socket/mod.rs
```diff
@@ -25,9 +25,194 @@ mod idleness;
...
+        Algorithms::from_bits_truncate(self.0 >> 1)
+    }
+
+    pub(crate) const fn preferred(self) -> Algorithms {
+        // `supported` bits would be discarded.
+        Algorithms::from_bits_truncate(self.0)
+    }
+}
+
+/// Layout:
```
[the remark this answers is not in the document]
nerodono: >This (and the same for compression) is internal implementation, so it's strange to use public comments for it, it should be a regular comment.

IMO it should be here in this form for two reasons:
- These items aren't public, so doc-comments wouldn't be seen by the user, but by the person who writes the `elfo-network` code (popped up in window while typing/hovering or something)
- `cargo doc` has `--document-private-items` key, so, it actually can appear in documentation if you wish to

> asciiflow.com

Sure, I'll do that

elfo-network/src/socket/mod.rs
nerodono: Like in the comment stated above, I misunderstood the intention. Gonna rewrite that.

elfo-network/src/config.rs
```diff
@@ -58,22 +58,46 @@ pub struct Config {
     pub idle_timeout: Duration,
 }
 
+/// Preference.
+#[derive(Debug, Clone, Copy, Deserialize)]
+#[serde(rename_all = "snake_case")]
```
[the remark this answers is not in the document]
nerodono: Got rid of this

elfo-network/src/socket/mod.rs
nerodono: Fixed. Resolves the initial concerns

elfo-network/src/socket/mod.rs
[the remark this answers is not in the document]
nerodono: Done

elfo-network/src/config.rs
loyd: Still "algorithms"

elfo-network/src/socket/capabilities/compression.rs
```diff
@@ -0,0 +1,157 @@
...
+        // 0 1 0 1 | Supported
+        // 1 0 1 0 | Preferred
+        // -------
+        // 1 1 1 1
+        let joined = supported | (preferred << 1);
+
+        Self(joined)
+    }
+
+    pub(crate) fn toggle(&mut self, algos: Algorithms, pref: Preference) {
```
loyd, line 70: Actually, It's not a "toggle", because this code does not work as expected for `toggle(Preferred); toggle(Disabled)` sequence.

loyd:
> First of all, this algorithm implies there's node which "leads" in choice of algorithm ("first node"), in the current implementation, lead node is node that accepts the connection.

What does it mean? Both peers perform the same intersection process.

elfo-network/src/discovery/mod.rs
```diff
@@ -135,12 +135,21 @@ impl Discovery {
...
+
+        let mut compression = socket::Compression::empty();
+        let cfg = &self.cfg.compression.algorithms;
+
+        compression.toggle(Algos::LZ4, cfg.lz4);
+
+        compression
+    }
+
     fn get_capabilities(&self) -> socket::Capabilities {
```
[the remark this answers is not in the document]
loyd: I see initial capabilities, not the resulting ones ("after a connection is established"), that's more important

elfo-network/src/config.rs
loyd: gonna fix on my own

PR #159, opened 2025-05-05
feat: implement `AnyMessage::downcast_mut()`

GoldsteinE:
Complements existing `AnyMessage::downcast()` and `AnyMessage::downcast_ref()`, allowing to modify stored `AnyMessage`s without reallocation.

### DIFF
--- a/elfo-core/src/message/any.rs
+++ b/elfo-core/src/message/any.rs
@@ -77,6 +77,15 @@ impl AnyMessage {
         &self.0.cast::<MessageRepr<M>>().as_ref().data
     }
 
+    /// # Safety
+    ///
+    /// The caller must ensure that the message is of the correct type.
+    pub(super) unsafe fn as_real_mut<M: Message>(&mut self) -> &mut M {
+        debug_assert_ne!(M::_type_id(), Self::_type_id());
+
+        &mut self.0.cast::<MessageRepr<M>>().as_mut().data
+    }
+
     /// Returns [`AnyMessageRef`] that borrows the message.
     pub fn as_ref(&self) -> AnyMessageRef<'_> {
         // SAFETY: `self` is valid for reads.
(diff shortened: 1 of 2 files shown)


elfo-core/src/message/any.rs
```diff
@@ -116,6 +125,24 @@ impl AnyMessage {
         M::_from_any_ref(self)
     }
 
+    /// Tries to downcast the message to a mutable reference to the concrete type.
+    ///
+    /// Note: it returns `Some(&mut self)` if `M` is [`AnyMessage`].
+    pub fn downcast_mut<M: Message>(&mut self) -> Option<&mut M> {
```
loyd, line 131: `#[inline]`?

elfo-core/src/message/any.rs
```diff
@@ -116,6 +125,24 @@ impl AnyMessage {
...
+        self.is::<M>()
+            // SAFETY: `self` is of type `M`, checked above.
+            .then(|| unsafe { self.downcast_mut_unchecked() })
+    }
+
+    /// # Safety
+    ///
+    /// The caller must ensure that the message is of the correct type.
+    pub(crate) unsafe fn downcast_mut_unchecked<M: Message>(&mut self) -> &mut M {
+        // If `M != AnyMessage` then `as_real_ref()` is called.
```
loyd, line 141: `as_real_ref` -> `as_real_mut`

elfo-core/src/message/any.rs
GoldsteinE: Sure. While we’re at it, should `_unchecked()` versions also have `#[inline]`?

loyd: Most likely it doesn't matter until it's public

GoldsteinE: Okay. I added `#[inline]` to this one.

loyd (approved):

PR #163, opened 2025-06-02
feat(network): implement connection manager

nerodono:


elfo-network/src/discovery/mod.rs
```diff
@@ -27,20 +28,23 @@ mod diff;
 /// TODO: should be different for groups and actors.
 const INITIAL_WINDOW_SIZE: i32 = 100_000;
 
+#[message]
+enum EstablishSide {
```
loyd, line 32: I don't think that the terms "Server" and "Client" are right to use here. It's more about an initiator.

Anyway, `ConnId` should be used in all cases

nerodono: `Client` and `Server` terms are unfortunate, agreed.

> Anyway, ConnId should be used in all cases

There's a reason it's not used when someone initiates connections to us - this would require inserting to slotmap from the stream, which would require moving it inside.

nerodono: Renamed to `ConnectionAllocation`, IMO this name describes purpose better

elfo-network/src/connman.rs
```diff
@@ -0,0 +1,366 @@
...
+    type Output = Conn;
+
+    #[track_caller]
+    fn index(&self, index: ConnId) -> &Self::Output {
+        self.get(index).unwrap()
+    }
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, IsVariant)]
+pub(crate) enum State {
```
loyd, line 248: Why are so many internal details public?

elfo-network/src/connman.rs
```diff
@@ -0,0 +1,366 @@
...
+    ) -> ManagedConn<'_> {
+        let id = self.conns.insert(Conn {
+            role,
+            transport,
+            state: State::New,
+        });
+        self.task_queue
+            .push(Task::new(Instant::now(), TaskData::ConnectNew(id)));
+        // SAFETY: just inserted.
+        unsafe { self.get_mut(id).unwrap_unchecked() }
```
loyd, line 199: Let's use `expect()` because we don't have enough tests to detect regression, and safety guarantees are across several methods.

elfo-network/src/connman.rs
```diff
@@ -0,0 +1,366 @@
...
+slotmap::new_key_type! {
+    struct QueuePlace;
+}
+
+#[derive(Debug)]
+pub(crate) struct ConnMan {
+    config: config::Config,
+    conns: slotmap::SlotMap<ConnId, Conn>,
+
+    pub(crate) task_queue: Vec<Task>,
```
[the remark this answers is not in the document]
nerodono: It was accidentally left here for debugging purposes

elfo-network/src/connman.rs
```diff
@@ -0,0 +1,366 @@
...
+    }
+}
+
+pub(crate) struct StateTransition<'a, 't>(&'a mut ManagedConn<'t>);
+
+impl<'t> StateTransition<'_, 't> {
+    fn transition(&mut self, map: impl FnOnce(&mut ManagedConn<'t>, State) -> Option<State>) {
+        let prev = self.0.state;
+        if let Some(next) = map(&mut *self.0, prev) {
+            tracing::debug!(
```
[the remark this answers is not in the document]
nerodono: Still, gives some understanding about what's happening, especially if we're moving most lifecycle-related logs inside manager

elfo-network/src/connman.rs
```diff
@@ -0,0 +1,366 @@
...
+    }
+}
+
+#[derive(Debug, Clone, Copy, Display)]
+#[display("no such connection {_0:?}")]
+pub(crate) struct NoSuchConn(pub ConnId);
+
+slotmap::new_key_type! {
+    #[derive(Display)]
+    #[display("conn:{:X}", self.0.as_ffi())]
```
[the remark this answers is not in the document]
nerodono: The point here is simple - _even if_ id gets into logs, it would be easy recognizable (the `conn` part makes it easy to identify what is this exactly), and, since ID in base-10 is practically gibberish, base-16 is more appealing, since it's easy to spot "generation" and "slot" part, while leaving exact representation opaque, and if we, say, ever introduce `null`-keys, it would be spotted in display impl with ease. using `Debug` instance is OK, should I or I convinced you :smile: ?

elfo-network/src/discovery/mod.rs
```diff
@@ -126,6 +138,26 @@ impl Discovery {
...
+            info!(
+                after = ?advise.duration_since(Instant::now()),
+                "scheduled next connection opening",
+            );
+        }
+
+        let capabilities = self.get_capabilities();
+        for (prev_state, id) in connections {
+            let conn = self.connman.get_mut(id).unwrap();
+            info!(%conn, %prev_state, "opening connection");
```
nerodono, line 154: add capabilities here

elfo-utils/src/lib.rs
```diff
@@ -11,6 +11,35 @@ pub mod time;
...
+/// part of the pattern could be returned:
+/// ```
+/// use elfo_utils::assert_matches;
+///
+/// assert_matches!(1, 1);
+/// let x = assert_matches!((1, 2), x = (x, _));
+/// assert_eq!(x, 1);
+/// ```
+#[macro_export]
+macro_rules! assert_matches {
```
loyd, line 24: 1. `elfo-utils` for very basic code used in multiple elfo crates
2. Don't see a reason for such sugar now

elfo-network/src/connman.rs
```diff
@@ -0,0 +1,346 @@
...
+pub(crate) use self::config::Config;
+
+#[cfg(test)]
+mod tests;
+
+mod config;
+
+/// Helper type which suggests when new timer should be
+/// created.
+pub(crate) struct AdviseTimer {
```
loyd, line 27: This is not a part of the connection manager

elfo-network/src/connman.rs
loyd: It seems the only way to access `State` is through `make_connections()` call, innit? And.. only for logging purposes

elfo-network/src/connman.rs
```diff
@@ -0,0 +1,346 @@
...
+    }
+
+    pub(crate) const fn state(&self) -> State {
+        self.state
+    }
+
+    fn switch_state(&mut self, new: State) -> State {
+        let prev = std::mem::replace(&mut self.state, new);
+        debug!(
+            from = %prev,
```
loyd, line 262: Still useless log. I prefer to see one log per state switch with info about peer, socketinfo, prev/new status

elfo-network/src/connman.rs
```diff
@@ -0,0 +1,346 @@
...
+                id,
+                conn,
+                config: &self.config,
+            })
+            .ok_or(NoSuchConn(id))
+    }
+
+    /// Reallocate ID for that connection, previous id would be
+    /// no longer accessible.
+    pub(crate) fn realloc_id(&mut self, prev: ConnId) -> Result<ManagedConn<'_>, NoSuchConn> {
```
loyd, line 209: Why is it public?

elfo-network/src/discovery/mod.rs
```diff
@@ -126,6 +138,26 @@ impl Discovery {
         Ok(())
     }
 
+    fn make_connections(&mut self) {
+        let (next_check_advise, connections) = self.connman.make_connections();
+        if let Some(advise) = next_check_advise.and_then(|adv| self.advise_timer.feed(adv)) {
```
loyd, line 143: `next_check_advise = None` means "do not schedule", innit? Now it's possible to trigger a spurious tick

elfo-network/src/discovery/mod.rs
```diff
@@ -27,20 +30,26 @@ mod diff;
 /// TODO: should be different for groups and actors.
 const INITIAL_WINDOW_SIZE: i32 = 100_000;
 
+#[message]
+struct MakeConnections;
```
loyd, line 34: (minor, naming)

https://actoromicon.rs/ch03-03-sources.html#delays

> adhering to best practice by using the 'Tick' suffix

Also, `Make` does not obviously mean "Open"

elfo-network/src/discovery/mod.rs
```diff
@@ -126,6 +138,26 @@ impl Discovery {
...
+                "scheduled next connection opening",
+            );
+        }
+
+        let capabilities = self.get_capabilities();
+        for (prev_state, id) in connections {
+            let conn = self.connman.get_mut(id).unwrap();
+            info!(%conn, %prev_state, "opening connection");
+
+            let stream = Self::open_existing_connection(capabilities, &self.node_map, conn);
```
loyd, line 156: (minor, naming)
We can call it "existing" only if the connection ID remains the same, but it doesn't.

Cargo.toml
```diff
@@ -35,6 +35,9 @@ toml = "0.8.14"
 bytesize = { version = "2", features = ["serde"] }
 proptest = "1.6"
 criterion = "0.5.1"
+slotmap = "1.0.7"
+
+elfo-utils = { version = "0.2.6", path = "./elfo-utils" }
```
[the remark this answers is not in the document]
nerodono: Actually, let's keep things as they were

elfo-network/src/connman.rs
nerodono: > How?

through the state transitions, `t.failed(..)`, `t.accepted(..)`, etc, partly exposes what possible states are. Encapsulated `State`, nonetheless.

elfo-network/src/connman.rs
nerodono: No reasons to hide it, can be used later from outside of connection manager, but let's make it private before that

elfo-network/src/discovery/mod.rs
nerodono: now it's unneeded here

elfo-network/src/discovery/mod.rs
loyd: Ok, right now there are no spurious ticks. However, the connection manager will also solve situations when an incoming connection cancels the timer, in the case when this incoming connection refers to the same groups that's scheduled:
1. Open a connection for `(a, b)` group pair => failed => set timer up
2. Accept a connection also for `(a, b)` group pair => cancel timer tick

elfo-network/src/connman.rs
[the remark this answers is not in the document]
loyd: Hex is ok, my point is that we don't need any `"conn:"` prefix here

elfo-network/src/connman.rs
```diff
@@ -0,0 +1,408 @@
...
+        self.transition(|this, prev| {
+            if prev.is_failed() {
+                return None;
+            }
+
+            let recon_interval = this.config.reconnect_interval;
+            let now = Instant::now();
+            let reconnect_at = now.checked_add(recon_interval).unwrap();
+
+            Some(StateInner::Failed { reconnect_at })
```
loyd, line 376: Is it intentional to leave `socket_info` unchanged?

nerodono: It is, as long as connection is not re-established, IMO better to keep that info until re-establish.

elfo-network/src/discovery/mod.rs
nerodono: I guess spurious tick in that case is OK, it shouldn't be a problem, and anyway, `AdviseTimer` could keep `Delay` as well to terminate it when necessary. I don't see why it's a problem.

elfo-network/src/connman.rs
nerodono: Erased

elfo-network/Cargo.toml
```diff
@@ -38,6 +38,7 @@ bitflags = "2.3.2"
 lz4_flex = { version = "0.11.1", default-features = false, features = ["std"] }
 byteorder = "1.4.3"
 turmoil06 = { package = "turmoil", version = "0.6", optional = true }
+slotmap = { workspace = true, features = ["serde"] }
```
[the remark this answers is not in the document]
nerodono: Sadly, no, `new_key_type` just creates wrapper around `KeyData`, which is not `Serialize` or `Deserialize`.

elfo-network/src/discovery/mod.rs
```diff
@@ -198,81 +227,89 @@ impl Discovery {
...
-            socket = %socket.info,
-            peer = %socket.peer,
-            role = %msg.role,
-            capabilities = %socket.capabilities,
-        );
+        log_conn!(debug, conn, "new connection established");
+        let role = conn.role();
+        let id = conn.id();
 
         if socket.peer.node_no == self.node_map.this.node_no {
```
loyd, line 311: What's the problem with moving this logic of decision-making into the connection manager?

elfo-network/src/discovery/mod.rs
```diff
@@ -334,15 +355,21 @@ impl Discovery {
                     // TODO: check launch_id.
                 }
 
-                self.control_maintenance(socket, msg.transport.clone());
+                self.control_maintenance(socket, id);
+
+                let mut conn = self.connman.get_mut(id).unwrap();
+                conn.change_state(|t| t.accepted(ConnectionRole::Control));
+                log_conn!(info, conn, "new connection accepted");
```
loyd, line 362: same

elfo-network/src/discovery/mod.rs
```diff
@@ -387,10 +414,15 @@ impl Discovery {
...
-                    transport: msg.transport.clone(),
-                };
+                let mut conn = self.connman.get_mut(id).unwrap();
+                conn.change_state(|t| {
+                    t.accepted(ConnectionRole::Data {
+                        local_group_no,
+                        remote_group_no,
+                    })
+                });
+                log_conn!(info, conn, "new connection accepted");
```
loyd, line 424: same

elfo-network/src/discovery/mod.rs
```diff
@@ -126,6 +127,32 @@ impl Discovery {
...
+            info!(
+                after = ?advise.duration_since(Instant::now()),
+                "scheduled next connection opening",
+            );
+        }
+
+        let capabilities = self.get_capabilities();
+        for (prev_state, id) in connections {
+            let conn = self.connman.get(id).unwrap();
+            log_conn!(
```
loyd, line 143: Logged by `make_connections` due to switch to `Establishing`?

elfo-network/src/discovery/mod.rs
loyd: Still don't see what the problem is with terminating the timer if `None` is returned to avoid any potential spurious ticks, avoiding misleading logs, dumps, and metrics (when you think that something happens even when it doesn't)

nerodono: Returning `None` doesn't means "no timer should be active", `None` means "previous timer is still ticking". Redesigned `feed(..)` return type to describe this

elfo-network/src/discovery/mod.rs
nerodono: It could be, moved.

elfo-network/src/discovery/mod.rs
loyd: > Returning None doesn't means "no timer should be active"

The code tells the opposite

nerodono: Well, how so? I agree that option in `make_connections(..) -> Option<..>` means `no timer should be active`, but `feed(..) -> Option<..>` had entirely different meaning - `None = there's no need to set up new timer, since previous is already active`. Anyway, new type describes that invariant.

PR #167, opened 2025-07-10
feat: add reason to Terminate

Erigara:


elfo-core/src/messages.rs
```diff
@@ -81,12 +83,41 @@ pub struct ConfigUpdated {
 #[non_exhaustive]
 pub struct Terminate {
     pub(crate) closing: bool,
+    pub reason: Option<TerminateReason>,
```
loyd, line 86: Why do we need `Option` if `TerminateReason` has the `Unknown` variant?

Erigara: We don't, initially i thought that `Option<TerminationReason>` without `Unkown` would be a way to go, but i like flatten struct more.

Issue #132, opened 2024-07-03
Preserve routed messages if closed actor is supposed to restart


loyd:

qwfy:
Hi, thanks for the library, I tried several other actor systems (ractor and coerce), and found this one to be the best suited for my purpose.

I'm building a system with this library, and found that I need this feature for at-least-once delivery semantics, can you share some insights on this? (eg. what it is, what are the design considerations, how will it be implemented, will the mailbox survive operating system process restarts, things like that).

Curiously enough I used to work with Erlang and don't remember that I needed such feature.

loyd:
>  found this one to be the best suited for my purpose

Thank you for your interest in the project! Feel free to ask anything; I know that the documentation is the weak point of this project for now. And I hope you use the recent alpha releases because v0.1.0 is very outdated and limited by functionality.

The issue name is confusing. I'll try to explain what this issue is about. It's probably not what you are looking for, so please feel free to describe your use case, and I'll try to help you.

---

Let's consider the following example: GRPC clients want to subscribe to some data from a database with parameters. To reduce data transfer from the database, all subscriptions with the same parameters are handled by one specific streamer actor (for these parameters).
![Image](https://github.com/user-attachments/assets/4a35de3a-22e0-4770-9fe3-47213a021cc3)

However, it's pointless to transfer data if nobody is interested in data, so if all GRPC clients with specific parameters are disconnected, the specific streamer should be terminated. It can occur immediately if re`WATCH`ing is cheap or after some time if not (for instance, if the streamer reads data not from some specific DB but from some distributed system with some long process of initializing distributed subscriptions).

So, every new GRPC call with specific parameters leads to sending `SubscribeToX` to the streamer group. A new one is started inside the group if no actor handles streams with such parameters.

The problem is some race conditions if the streamer has decided to terminate, but a new `SubscribeToX` is sent while the termination process has not ended. More specifically, an actor no longer reads incoming messages but does not complete the termination process. In such a case, the grpc actor will get an error and should retry a subscription request (an additional boilerplate) or return an error to a client (an unnecessary error that is not the problem of the client).

[...]

Issue #68, opened 2022-05-30
`UpdateConfig` and `Ping` can stuck if an actor fails


loyd:
.. because a mailbox is destroyed only at restart moment.

PR #189, opened 2026-04-23
feat: mailbox single consumer

sargarass:
### Changed                                                                                                                            
  - **BREAKING** core/context: `recv`/`try_recv` now live on a move-only `MailboxConsumer` acquired once per spawn; cloned `Context`s are send-only.                                                                                             
  - core/mailbox: swap `tokio::Notify` → `DiatomicWaker` and `MpscQueue::dequeue` → `dequeue_unchecked`; −5…−6.5% on  `one_to_one/send_{routed,direct}` at 4+ producers, 1p1c unchanged except `send_direct` +4.6% (Xeon Gold 6526Y, SNC2). 
  
  See https://gist.github.com/sargarass/362a9cc3c7d1722eef50e979b92041be

### DIFF
--- a/elfo-core/src/actor.rs
+++ b/elfo-core/src/actor.rs
@@ -15,9 +15,10 @@ use crate::{
     envelope::Envelope,
     errors::{SendError, TrySendError},
     group::TerminationPolicy,
-    mailbox::{Mailbox, RecvResult, config::MailboxConfig},
+    mailbox::{Mailbox, MailboxConsumer, config::MailboxConfig},
     messages::{ActorStatusReport, Terminate},
     msg,
+    object::{MappedOwnedObject, OwnedObject},
     request_table::RequestTable,
     restarting::RestartPolicy,
     scope,
@@ -104,6 +105,9 @@ impl ActorStartCause {
         matches!(self, ActorStartCause::OnMessage)
     }
 }
+
+/// A [`MailboxConsumer`] projected out of an actor's [`OwnedObject`].
+pub(crate) type OwnedMailboxConsumer = MailboxConsumer<MappedOwnedObject<Mailbox>>;
 
 // === Actor ===
 
@@ -198,12 +202,11 @@ impl Actor {
         Some(envelope)
     }
 
-    pub(crate) async fn recv(&self) -> RecvResult {
-        self.mailbox.recv().await
-    }
-
-    pub(crate) fn try_recv(&self) -> Option<RecvResult> {
-        self.mailbox.try_recv()
+    /// Panics if a consumer is already attached to this actor's mailbox.
+    pub(crate) fn make_consumer(entry: OwnedObject) -> OwnedMailboxConsumer {
+        let mailbox =
+            MappedOwnedObject::map(entry, |obj| &obj.as_actor().expect("actor object").mailbox);
+        MailboxConsumer::new(mailbox)
     }
 
     pub(crate) fn request_table(&self) -> &RequestTable {
@@ -258,9 +261,12 @@ impl Actor {
         drop(control);
 
         if status.kind().is_finished() {
-            self.close();
-            // Drop all messages to release requests immediately.
-            self.mailbox.drop_all();
+            // No-op while the mailbox consumer is alive
+            if !self.mailbox.close_and_try_drain(scope::trace_id()) {
+                error!(
+                    "mailbox cannot be drained after termination, a context outlives the actor; pending requests may hang"
+                );
+            }
             self.finished.set();
         }
 
--- a/elfo-core/src/context.rs
+++ b/elfo-core/src/context.rs
@@ -14,7 +14,7 @@ use elfo_utils::unlikely;
 
 use crate::{
     ActorStatusKind,
-    actor::{Actor, ActorStartInfo},
+    actor::{Actor, ActorStartInfo, OwnedMailboxConsumer},
     actor_status::ActorStatus,
     addr::Addr,
     address_book::AddressBook,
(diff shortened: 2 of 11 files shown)


elfo/tests/restarting.rs
```diff
@@ -140,7 +140,9 @@ impl Drop for GuardedMessage {
 // See #68.
 #[tokio::test(start_paused = true)]
 async fn mailbox_must_be_dropped() {
```
loyd, line 142: This test stopped checking what's stated in #68 (not now, after https://github.com/elfo-rs/elfo/pull/118). Let's update this test by configuring the restart policy

elfo/tests/restarting.rs
```diff
@@ -140,7 +140,9 @@ impl Drop for GuardedMessage {
...
-    let blueprint = ActorGroup::new().exec(|_ctx| async {
+    let blueprint = ActorGroup::new().exec(|ctx| async move {
+        // Hold `ctx` until the actor finishes; its drop closes the mailbox.
+        let _ctx = ctx;
```
loyd: We faced the big question here. It means that `impl Drop for Context` leads to closing the mailbox, which makes impossible (?) to implement #132

elfo-core/src/object.rs
```diff
@@ -33,6 +35,49 @@ pub(crate) type BorrowedObject<'g> = BorrowedEntry<'g, Object>;
...
+        Self { entry, projected }
+    }
+}
+
+impl<P> Deref for MappedOwnedObject<P> {
+    type Target = P;
+
+    #[inline]
+    fn deref(&self) -> &P {
+        // SAFETY: `entry` keeps object alive, so `projected` remains valid.
```
loyd: One invariant is missing here: `OwnedObject` contains pinned memory

elfo-core/src/context.rs
```diff
@@ -995,10 +993,25 @@ impl<C, K> Context<C, K> {
...
         self.stats = Stats::startup();
+        self.actor = Some(MappedOwnedObject::map(owned, |obj| {
+            obj.as_actor().expect("actor object")
+        }));
+        self
+    }
+
+    /// # Safety
+    ///
+    /// At most one consumer may be acquired per actor's mailbox.
```
loyd: I think it's better to add a flag (either in the control block or `rx_waker` cacheline) to ensure it

elfo-core/src/context.rs
```diff
@@ -797,18 +795,16 @@ impl<C, K> Context<C, K> {
             self.pre_recv().await;
 
             let envelope = 'received: {
-                let actor = ward!(
-                    self.actor.as_ref().and_then(|o| o.as_actor()),
-                    return Err(TryRecvError::Closed)
-                );
+                let consumer = ward!(self.consumer.as_mut(), return Err(TryRecvError::Closed));
```
loyd, line 798: It seems that `self.consumer` and `self.stage` are implicitly related now? Let's think about combining them and doing something like `let consumer = self.pre_recv().await;`.

Not sure whether it will be concise, but it makes sense to try.

elfo/tests/restarting.rs
sargarass: Preserved old cleaning behavior in sv to not block [#132](https://github.com/elfo-rs/elfo/issues/132)

elfo-core/src/context.rs
sargarass: Because `pre_recv` needs to be called every time in the loop close to mailbox & sources, it is not easy to extract (consumer, sources, stage) without getting code ugly

elfo-core/src/object.rs
sargarass: fixed

elfo/tests/restarting.rs
sargarass: fixed (added policy OnFailures)

elfo-core/src/context.rs
sargarass: Fixed

elfo-core/src/actor.rs
```diff
@@ -258,9 +261,8 @@ impl Actor {
         drop(control);
 
         if status.kind().is_finished() {
-            self.close();
-            // Drop all messages to release requests immediately.
-            self.mailbox.drop_all();
+            // No-op while the consumer is alive
```
loyd: It seems no one will drop messages if a context is moved and lives longer than the corresponding task

elfo-core/src/mailbox.rs
```diff
@@ -218,80 +229,122 @@ impl Mailbox {
...
+    /// `MailboxConsumer` is attached.
+    #[cold]
+    pub(crate) fn close_if_unattached(&self, trace_id: TraceId) -> bool {
+        let mut control = self.control.lock();
+        if control.consumer_attached {
+            return false;
+        }
+        let _ = self.close_inner(&mut control, trace_id);
+        // SAFETY: No other consumer is exist at this time.
+        unsafe { while self.queue.dequeue_unchecked().is_some() {} }
```
loyd: I think it's cleaner to make `MailboxConsumer` here (`MailboxConsumer::new()` should return `Option<Self>` for it).

Even better to call `MailboxConsumer::try_recv()` in the loop because it also releases the semaphore without waiting until the corresponding `Mailbox` instance is dropped.

elfo-core/src/actor.rs
loyd: Also, detecting such a situation means we can also `warn!` about it, because keeping the context after the actor terminates is not the right way

sargarass: added `error!` to highlight this case + drop cleanup in mailbox consumer + test

elfo-core/src/mailbox.rs
sargarass: We don't want the `try_recv` overhead here (a `tx_semaphore.add_permits(1)` per envelope); the semaphore is already closed, so the permits can't be acquired and there are no waiters to wake.

loyd (approved):

